PR #8, opened 2021-09-29
Added options to apply default style to .ass subtitles and set subtitle back color

Alpacalord:
Used mpv's settings sub-ass=no to make mpv use user settings
Also used mpv's sub-back-color=color to make mpv render a background behind the subtitles.


Alpacalord:
Mistake in ui, subs ASS override was visible on any settings page. Will fix later today.

PR #9, opened 2021-09-29
Added options to apply default style to .ass subtitles and set a backcolor

Alpacalord:
Used mpv's settings sub-ass=no to make mpv use default style for ass subtitles (now without a bugged button)
Used mpv's sub-back-color=color to make mpv render a background behind the subtitles.


cmdrkotori:
Thanks.

PR #46, opened 2022-08-23
Added lang group box.

xRoBoTx:
Added language change group box in player settings. I hope full functionality would be implemented later.
Corrected hyperlink mistake in russian translation.

### DIFF
--- a/settingswindow.cpp
+++ b/settingswindow.cpp
@@ -508,19 +508,12 @@ void SettingsWindow::generateVideoPresets()
 {
     SettingMap videoWidgets;
 
-    // Deprecated way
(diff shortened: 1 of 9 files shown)


cmdrkotori:
LGTM. I'm not sure what's going on in the Russian translation but I don't use that.
Speaking of translations, I lost access to transifex when my email address changed. So we'll have to look for some alternative.

Regarding the settingswindow.cpp changes around line 500, I see you deleted the old way of inserting items and uncommented the non-deprecated way introduced in Qt 5.15. We can probably uncomment+switch the rest of these occurrences to the new way now that Qt 5.15 is in the latest ubuntu.

PR #61, opened 2023-02-13
Qt6

cmdrkotori:
Initial fixes and enhancements:
* Got rid of a static cast in qthelper.cpp
* i18n files are now embedded as resources

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,13 @@ Suggestions welcome.
 
 ### Prerequisites
 
-You need the Qt5 SDK installed and a recent edition of libmpv.  On Ubuntu you
-can usually install the required libraries with the ``qtcreator``,
-``qt5-default``, ``libqt5x11extras5-dev``, ``qttools5-dev-tools``, 
-``qtbase5-private-dev`` and ``libmpv-dev`` packages.  You may also need to install
-``git``, ``build-essential`` and ``pkg-config``.  A recent edition of [libmpv]
-means either from git head or at least version 0.29.0.  The mpv options for this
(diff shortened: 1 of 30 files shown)


cmdrkotori:
This needs testing. As this initial port was written on a Linux laptop while on holidays, the platform-specific components (Windows and Mac) still need an appropriate compiler to check for needed changes. Also there may be peculiarities that are present on non-Arch distros. All of that needs to wait until I get back and have time to test.

ahjolinna:
have you thought about moving to cmake with Qt6?  thats what they are pushing/moving towards as they kinda abandoned qmake https://doc.qt.io/qt-6/qt6-buildsystem.html

I do get that this is just initial port, and I'm currently testing it

cmdrkotori:
> have you thought about moving to cmake with Qt6? thats what they are pushing/moving towards as they kinda abandoned qmake https://doc.qt.io/qt-6/qt6-buildsystem.html
> 
> I do get that this is just initial port

I think I'll be using qmake as long as I can because I much prefer its syntax over cmake. Yes I have used both; it's a taste issue. :+1:

cmdrkotori:
OK, all seems in order. I'll merge this in a day or so if there aren't any complaints.

By the way, this will drop support for anything less than Windows 10.

ahjolinna:
seems to be working it at least builded for me on OBS for openSUSE tumbleweed: https://build.opensuse.org/package/show/home:ahjolinna/mpc-qt

and it did run just fine on my (AMD) laptop with (KDE-git) wayland

PR #85, opened 2023-07-20
Window management + Fullscreen fixes

cmdrkotori:
This PR scratches a long standing itch that the window managing code was all over the codebase. Most (all?) window management code that was in main.cpp is now in its own module in a much more generic form. I had to keep some specifics to this project though, so it pulls in the mainwindow.h header.

It also implements what is required to enable "Fullscreen Monitor" control group inside the settings window by adding custom combo box control that tracks available monitors and using it in the settings window; moving the fullscreen switching code into the new window management module and making it work with an arbitrary monitor.

As I lack a second monitor, I tested this on a Linux Virtual Machine with "two monitors". It seems to work well. It might work on Windows too, but I can't test for that as a VirtualBox Windows VM lacks proper opengl support.


cmdrkotori:
Got a second monitor out of storage and wired it up. Works on my Windows (virtual) machine. 

HOWEVER I think this may have a bug on Windows when using two monitors of exactly the same model number. The second one will always be selected. We will need some way to differentiate between the two in the combo box. How to do this in a simple manner needs more thought. Perhaps we can copy the way X11 represents their geometry, and display that too.

cmdrkotori:
LGTM. Merging.

PR #99, opened 2023-11-28
Remove flatpak-isms from the project

cmdrkotori:
Fixes a bug I had during testing of Plasma 6, where pinning the icon to the taskbar resulted in an unlaunchable icon, or, if the icon was pinned a different way, a double icon. This bug is also present on Plasma 5, where launching a pinned icon to the taskbar results in the app launching under a second icon.

This only happens after the desktop file was renamed, and doesn't occur with any other application, so something in here was being done wrong.  Therefore, revert this to an earlier working version by removing all mentions of flatpak-isms from the project.

This stuff will have to be bundled on the flatpak repo, should we ever get to that.


cmdrkotori:
Windows CI broke again. Works on my machine though.

PR #136, opened 2024-08-11
Move most generated files to subdirectories

tbertels:
objects, moc and uic intermediate files to . subdirectories + executable to bin subdirectory
See https://doc.qt.io/qt-6.2/qmake-variable-reference.html

### DIFF
--- a/mpc-qt.pro
+++ b/mpc-qt.pro
@@ -13,6 +13,11 @@ TEMPLATE = app
 
 CONFIG += c++14
 
+DESTDIR=bin
+OBJECTS_DIR=.obj
+MOC_DIR=.moc
+UI_DIR=.ui
+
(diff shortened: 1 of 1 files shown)


cmdrkotori:
I wouldn't be too concerned about the MSYS2 CI, the MSYS2 build action breaks for seemingly no reason on occasion.

cmdrkotori:
LGTM. I can't check the effect it has on Windows build scripts as I'm away for two weeks and only have a Linux laptop on hand, but I'll check that when I get back.

Note: this appears to break the AUR package mpc-qt-git, but I'll let them them know that the build process has subtly changed.

cmdrkotori:
Good news, the AUR package didn't break. :tada:

PR #138, opened 2024-08-19
Remember video position

### LINKED ISSUE #44: Remembering video position?
Remembering the where you were upto last time you played the file?

I don't know if this is already a feature and I'm just dumb...

But it would be nice to be able to auto-start playback at the last played position in the file... if possible.

Love the program so far, seems like it will be a nice replacement for mpc-hc for me!

tbertels:
Fixes #44
Saves the position in recents and reuses it when opening a file from the file manager.
Stopping (not pausing) a video resets the position to the start.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -749,6 +749,10 @@ void Flow::setupFlowConnections()
     // manager -> this
     connect(playbackManager, &PlaybackManager::nowPlayingChanged,
             this, &Flow::manager_nowPlayingChanged);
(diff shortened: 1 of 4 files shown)


tbertels:
I completely forgot to add an option for this.

cmdrkotori:
Looks good. Remembering recent playback position is a common option included in sample mpv configs, so it's safe to say that most people would be okay with this.  So while an option would be nice (and would probably default to enabled), it's not really needed.

PR #139, opened 2024-08-20
Add a setting for Remember file position

tbertels:
Enabled by default.
The position is still saved when disabled, but the position isn't resumed.
For real privacy, the "Keep history of recently opened files" option should be disabled instead (and disables this too), but it doesn't do anything right now.


tbertels:
How do we update the .ts files?

cmdrkotori:
>How do we update the .ts files?

Change this setting in QtCreator
Screenshot_20240820_180201

and run this command
Screenshot_20240820_180220

cmdrkotori:
LGTM

PR #141, opened 2024-08-20
Add a command to show/hide the controls in fullscreen

tbertels:
Instead of moving the mouse, just pressing i (by default) shows/hides the controls while in fullscreen.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -601,10 +601,8 @@ void Flow::setupSettingsConnections()
             mainWindow, &MainWindow::setFullscreenOnPlay);
     connect(settingsWindow, &SettingsWindow::fullscreenExitAtEnd,
             mainWindow, &MainWindow::setFullscreenExitOnEnd);
-    connect(settingsWindow, &SettingsWindow::hideMethod,
-            mainWindow, &MainWindow::setBottomAreaBehavior);
-    connect(settingsWindow, &SettingsWindow::hideTime,
-            mainWindow, &MainWindow::setBottomAreaHideTime);
+    connect(settingsWindow, &SettingsWindow::fullscreenHideControls,
+            mainWindow, &MainWindow::setControlsInFullscreen);
     connect(settingsWindow, &SettingsWindow::hidePanels,
             mainWindow, &MainWindow::setFullscreenHidePanels);
     connect(settingsWindow, &SettingsWindow::volumeMax,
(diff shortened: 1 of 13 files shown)


tbertels:
Note that this is something that mpc-hc and mpc-be have always been missing as far as I know.

tbertels:
I didn't realize that this was shown in the View menu. 😅

cmdrkotori:
Excellent, however if I move the mouse the controls disappear.

There is a similar option in the settings window, in the fullscreen page.  How do you propose that this option would interact with it?

>I didn't realize that this was shown in the View menu. 😅

Don't worry too much about that.  If they don't see it they won't know about it.

tbertels:
> There is a similar option in the settings window, in the fullscreen page. How do you propose that this option would interact with it?

Great question actually!

After thinking about it, the best thing would probably be to link in inverse that new "Controls in Fullscreen" menu toggle to the "Hide controls in fullscreen" option.

However, on the subject of linking, I don't see why "Hide docked panels" should be disabled when "Hide controls in fullscreen" is unchecked.
The way I see it is that "Controls in Fullscreen" should only show the controls, not the Playlist, unless the user chooses to show it anyway by unchecking "Hide docked panels".

About the implementation, I'm thinking about moving the logic of these options to a MainWindow method. It kinda makes more sense for it to be there anyway.
Instead of the two hideMethod and hideTime signals, there would just be one with the two options in parameter.
And the hidePanels signal would left alone with a single call instead of a conditional call.

About calls, I've noticed by chance that Flow::settingswindow_settingsData gets called twice whenever we save settings. Is there any valid reason?

cmdrkotori:
> > There is a similar option in the settings window, in the fullscreen page. How do you propose that this option would interact with it?
> 
> Great question actually!
> 
> After thinking about it, the best thing would probably be to link in reverse that new "Controls in Fullscreen" menu toggle to the "Hide controls in fullscreen" option.
> 
> However, on the subject of linking, I don't see why "Hide docked panels" should be disabled when "Hide controls in fullscreen" is unchecked. The way I see it is that "Controls in Fullscreen" should only show the controls, not the Playlist, unless the user chooses to show it anyway by unchecking "Hide docked panels".

Yeah, when I tested how this code worked, I felt that hide docked panels would have to be changed.

> About the implementation, I'm thinking about moving the logic of these options to a MainWindow method. It kinda makes more sense for it to be there anyway. Instead of the two hideMethod and hideTime signals, there would just be one with the two options in parameter. And the hidePanels signal would left alone with a single call instead of a conditional call.

Seems good to me.  If you can simplify it a bit then go ahead. A bit of the reason for its current complexity is that it has to reparent itself under certain conditions.
 
> About calls, I've noticed by chance that Flow::settingswindow_settingsData gets called twice whenever we save settings. Is there any valid reason?

No reason, I thought it was only done once.

cmdrkotori:
Tested. All works here.

cmdrkotori:
Duplicate settings data should be fixed in 0e84a7fef2a2eb7bc40618e63a2b3816535503bd.

Issue #137, opened 2024-08-17
Next Video In Folder


RedMaxxss:
One feature of MPC-HC that I can't get to work in any player is that when I press PG Up-Dn, it starts playing the next video in the folder where the first video is.

Example: In the same folder, I have video files “A”, “B”, “C”, and I start playing B. When I press PG Dn, it will switch to video C.

abd3lraouf:
This is the first feature I'm missing now

tbertels:
See https://github.com/mpc-qt/mpc-qt/issues/74#issuecomment-1598168954 to enable it.

tbertels:
There are two settings which overlap:
* "Open next/previous file in folder when there is only one item in the playlist" in Tweaks
![image](https://github.com/user-attachments/assets/c5fe9cd6-2e75-4a28-9082-ec190d2ee4b1)

* "Default After Playback action:" "Play next in the folder" in Playback
![image](https://github.com/user-attachments/assets/7a4f611b-c356-4ecd-9219-53839fcd4b09)

Shouldn't "Open next/previous file in folder when there is only one item in the playlist" renamed to something like "Open manually next/previous file in folder when there is only one item in the playlist" and enabled by default?

So that by default, pressing Next or Previous (PgDown / PgUp by default) or clicking "Skip forward" or "Skip Backward" would play the next / previous file in folder, but would stop playing at the end of the file otherwise.

The auto play next file in folder would be set by "Play next in the folder" in Playback.

cmdrkotori:
IIRC the latter two options control the "After Playback" submenu available under the "Play" menu.  The "Default After Playback" setting decides what that submenu is set to when the player starts.

How it worked is that if the "Play After" submenu is set to "Play next in the folder", it will ALWAYS activate the "play the next file in the folder" code, even if there is something later on in the playlist, by replacing the active item in the playlist with the next file and beginning playback again. (the playlist will not advance.)

By default "Play After" is set to "Do Nothing" because the code finds the next file from the active playlist.  This copies the behavior from MPC-HC.  But perhaps it could be renamed to "Use Playlist" so that it reflects what's actually going on; MPC-HC didn't advertise its playlist functionality all that well.

On the other hand the "only one item in the playlist" option would only work if there's one file in the playlist.  The typical way to end up with single file playlists is to launch a single video file from the file manager.  So turning it on makes it much easier to binge watch entire seasons from the first episode in the file manager.

[...]

PR #144, opened 2024-08-24
Play next/previous video in folder

tbertels:
By default, PgDown and PgUp will play the next/previous video in folder if there's no more chapter and there's only one file in the playlist.

With the renamed "Play next file in folder when there is only one item in the playlist" option enabled, the next file in folder will play without pressing anything when the end of the file is reached and the playlist has only the current file.

Fixes issue #137


tbertels:
`PlaybackManager::checkAfterPlayback(bool playlistMode)` is always called with `playlistMode` set to `true`.
It used to be also called with `false`.
Could it be removed now?

tbertels:
I'm thinking about adding key shortcuts to open the next/previous file in folder even if there are more chapters after, I think MPC-HC has this.

manager.cpp
```diff
@@ -248,7 +248,7 @@ void PlaybackManager::navigateToNextChapter()
 {
     int64_t nextChapter = mpvObject_->chapter() + 1;
     if (nextChapter >= numChapters)
-        mpvw_playbackIdling();
+        playNext();
```
cmdrkotori, line 251: This change skips the check for decrementing the extra play times counter when navigating to past-the-end chapter.  If you wish to keep this, the extra play times counter needs to be set back to zero.

cmdrkotori:
> `PlaybackManager::checkAfterPlayback(bool playlistMode)` is always called with `playlistMode` set to `true`. It used to be also called with `false`. Now it seems like we're always in playlist mode, it just contains one file or more. Could it be removed?

Yes it could be removed.  There used to be some obscure mode where a file could be played without it landing in the playlist, but it's gone now.

> I'm thinking about adding key shortcuts to open the next/previous file in folder even if there are more chapters after, I think MPC-HC has this.

If there's something MPC-HC is doing that we aren't, I consider it a bug unless there's a good reason why we can't/don't.

> Maybe "Play next file in folder when there is only one item in the playlist" could be renamed to something like "Play automatically next file in folder when there is only one item in the playlist" to make it clearer that the next file will play... automatically.

I'm not against this. It is also possible to add a separate option (default on) for controlling whether automatic file progression is active or not.

manager.cpp
```diff
@@ -646,7 +646,7 @@ void PlaybackManager::updateSubtitleTrack()
     mpvObject_->setSubtitleTrack(subtitleEnabled ? subtitleTrackSelected : 0);
 }
 
-void PlaybackManager::checkAfterPlayback(bool playlistMode)
+void PlaybackManager::checkAfterPlayback()
```
cmdrkotori, line 649:
```
manager.cpp:820:27: error: no matching function for call to ‘PlaybackManager::checkAfterPlayback(bool)’
  820 |         checkAfterPlayback(true);
      |         ~~~~~~~~~~~~~~~~~~^~~~~~
```
Might want to amend this commit + force-push.

manager.cpp
cmdrkotori: Yes and yes.  For testing purposes each track can also be changed individually by the "Extra Play Times" submenu under "Edit".
And you're right, I forgot about that setting.

tbertels:
> > I'm thinking about adding key shortcuts to open the next/previous file in folder even if there are more chapters after, I think MPC-HC has this.
> 
> If there's something MPC-HC is doing that we aren't, I consider it a bug unless there's a good reason why we can't/don't.
> 
I've just checked it out: it opens the next/previous file in playlist or, if there's only one, the next file in folder with Ctrl+PgDown and Ctrl+PgUp by default. 

> > Maybe "Play next file in folder when there is only one item in the playlist" could be renamed to something like "Play automatically next file in folder when there is only one item in the playlist" to make it clearer that the next file will play... automatically.
> 
> I'm not against this. It is also possible to add a separate option (default on) for controlling whether automatic file progression is active or not.

By "automatic file progression", do you mean playing the next file in folder without user interaction?
Because MPC-HC doesn't do that, you have to press PgDown.

So I was thinking about an option to open the next file in folder manually (enabled by default) and an option to open it automatically (disabled by default).
Maybe "Use next file in folder when there is only one item in the playlist" and "Play automatically next file in folder when there is only one item in the playlist". I'm not sure that's clear enough.

cmdrkotori:
> By "automatic file progression", do you mean playing the next file in folder without user interaction?
> Because MPC-HC doesn't do that, you have to press PgDown.

I was speaking in general (both playlist and folder) in relation to the existing code.  But you're right MPC-HC doesn't do that, it should be changed to match.  Playlists without automatic playlist progression doesn't make sense anyway.

>So I was thinking about an option to open the next file in folder manually (enabled by default) and an option to open it automatically (disabled by default).
>
>Maybe "Use next file in folder when there is only one item in the playlist" and "Play automatically next file in folder when there is only one item in the playlist". I'm not sure that's clear enough though.

It's clear enough. You could also go with "Use next file in folder when there is only one item in the playlist" (enabled by default, as you said) and "Wait for navigation when there is only one item in the playlist" (enabled by default).  I'll leave this up to you.

tbertels:
> > So I was thinking about an option to open the next file in folder manually (enabled by default) and an option to open it automatically (disabled by default).
> > Maybe "Use next file in folder when there is only one item in the playlist" and "Play automatically next file in folder when there is only one item in the playlist". I'm not sure that's clear enough though.
> 
> It's clear enough. You could also go with "Use next file in folder when there is only one item in the playlist" (enabled by default, as you said) and "Wait for navigation when there is only one item in the playlist" (enabled by default). I'll leave this up to you.

MPC-HC actually uses the equivalent of this Playback option we talked about in the issue to enable or not the auto play next file in folder:

image

And it does it this way: first the files of the playlist, then the files in the folder.

MPC-QT does the same with the equivalent setting.
So I wonder if it wouldn't be better to just use that option instead of adding a second one which will do the exact same thing, now that the previous/next command feature will have its own option.

cmdrkotori:
>So I wonder if it wouldn't be better to just use that option instead of adding a second one which will do the exact same thing, now that the previous/next command feature will have its own option.

Yes it would indeed. In fact I noticed when writing the code for "default after playback" (added much later) that I was stomping on my own foot having a semi-related option in the tweaks page, so I had to write it such a way that certain actions were invoked by multiple options, certain options took priority, and so on.

cmdrkotori:
>I don't know if the users really want this though, as it can make them lose the playlist.

This is a concern, yes. I know of at least one mpv dev who uses mpc-qt to play their 10000-file-strong music collection that I presume is spread over multiple folders, so this is probably not desirable.  Adding so many files back would be a pain for them.

>If that's a real use case, I think it would be better to add a command to empty the playlist, which would allow continuing playing the files in the folder.

To support the existing use-case, there would have to be an option to turn it off I think.

tbertels:
An option to turn off the command?
Given the fact that the user can just open a different file to empty the playlist, maybe it's just better to let it that like it is right now and just rename the option "Play next in the folder" to let the user know that it works just like the next file and next chapter (once at the end) commands (playlist if not empty, otherwise folder).

cmdrkotori:
Yeah I'm fine with that.

mpvwidget.h
```diff
@@ -76,6 +76,7 @@ class MpvObject : public QObject
     void setVideoTrack(int64_t id);
     void setDrawLogo(bool yes);
     void setVolume(int64_t volume);
+    bool eofReached();
```
cmdrkotori, line 79: nitpick: this should be further down the page with the other getters, perhaps just before playLength().

cmdrkotori (approved):
LGTM. I'll do some testing and merge it in the morning if that's okay with you.

PR #148, opened 2024-08-29
Fix MSYS2 #2

tbertels:
Use auto line ending git config
Set the version to today's date to avoid error with windres


tbertels:
We'll see if it holds, but my theory is that windres doesn't like the commit sha as a version number.
That would explain the randomness.
The line endings change is related to https://github.com/marketplace/actions/setup-msys2#actionscheckout-and-line-endings

PR #150, opened 2024-08-30
Switch to C++17 and replace qAsConst with standard C++ feature

tbertels:
Switch to c++17 (gcc default since version 11).

Replace qAsConst with std::as_const
qAsConst is deprecated with Qt6, c++17 introduced std::as_const which does the exact same thing.


cmdrkotori:
Thanks, this has been on my to do list for a while.

PR #151, opened 2024-08-31
Add radio buttons in the Audio, Subtitles and Video Stream menus

### LINKED ISSUE #106: displaying a check mark next to the selected track in the menu
there is a missing check mark next to the subtitle tracks and audio, video streams that are currently selected for playback!
it would be nice if you added this.

[image](https://github.com/mpc-qt/mpc-qt/assets/14019010/549b09c4-529c-4caf-9bc1-a74577f95959)

image

what does this icon do, it doesn't have any description on hover?
image
upd 
I realized that she turns off subtitles, she lacks a description and some more understandable icon that these are subtitles.
https://cdn-icons-png.flaticon.com/512/565/565286.png

tbertels:
Adds radio buttons in the Audio, Subtitles and Video Stream menus, showing the selected one.
Because calling MpvObject::setMpvPropertyVariant while using radio buttons (instead of checkboxes) leads to strange behavior, we need to check the right button manually. Otherwise, two clicks are needed.

Fixes #106

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -718,6 +718,14 @@ void Flow::setupMpvObjectConnections()
     connect(mpvObject, &MpvObject::chaptersChanged,
             propertiesWindow, &PropertiesWindow::setChapters);
 
+    // mpvwidget -> mainwindow
+    connect(mpvObject, &MpvObject::audioTrackSet,
+            mainWindow, &MainWindow::audioTrackSet);
+    connect(mpvObject, &MpvObject::subtitleTrackSet,
+            mainWindow, &MainWindow::subtitleTrackSet);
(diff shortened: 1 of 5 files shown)


tbertels:
Note that using int64_t for the audio, subtitles and video stream ids isn't needed.
We could change them to use int and avoid the static_casts.

cmdrkotori:
>Note that using int64_t for the audio, subtitles and video stream ids isn't needed.
>We could change them to use int and avoid the static_casts.

A shorter form of this is to call `int()` constructor on the int_64 value instead of static_cast, like this:
`audioTracksGroup->actions()[int(id)-1]->setChecked(true);`

PR #152, opened 2024-08-31
Increase recents size

tbertels:
Set it to 20
Loop to remove excess files to support future setting to increase size


tbertels:
20 is the default setting of MPC-HC.
To add a setting to increase further the size (up to 1000 like MPC-HC?), the Clear list button should be put to the top (like MPC-HC).

PR #153, opened 2024-08-31
Use pointing hand cursor where needed

tbertels:
It's needed for the seek bar and the sliders in the settings.

### DIFF
--- a/settingswindow.ui
+++ b/settingswindow.ui
@@ -1370,6 +1370,9 @@ media file played</string>
                 <property name="tickInterval">
                  <number>10</number>
                 </property>
+                <property name="cursor">
+                 <cursorShape>PointingHandCursor</cursorShape>
+                </property>
                </widget>
               </item>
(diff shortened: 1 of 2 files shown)


tbertels:
About the seek bar: is there a way to extend the hover area to the area between the seek bar and the video (like in MPC-HC) to make it easier to use it?

Or maybe we could do like in MPC-BE and add that space to the seek bar. It would also make it possible to show the file name in it like in MPC-BE.

cmdrkotori:
>About the seek bar: is there a way to extend the hover area to the area between the seek bar and the video (like in MPC-HC) to make it easier to use it?

Yes it means changing these top/bottom margins to 0 and putting that padding in the code instead.

Screenshot_20240831_212247

cmdrkotori:
Fixed the seekbar in master. I went with the mpc-hc approach and made the padding clickable.

PR #156, opened 2024-09-02
Improve seekbar tooltip

tbertels:
Improve readability of parameters in MainWindow::position_hoverValue

Don't show chapter info in the seekbar tooltip when there are no chapters. No more " - <unknown>" in the tooltip.

Make the seekbar tooltip follow the playback time short mode setting in Tweaks. "00:00:00" instead of "00:00:00.000".
Note that MPC-HC had always had a bug where the chapter time was always shown in long mode.

Update the chapters when changing the playback time short mode option.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -456,6 +456,8 @@ void Flow::setupMainWindowConnections()
             playbackManager, &PlaybackManager::setAfterPlaybackOnce);
     connect(mainWindow, &MainWindow::afterPlaybackAlways,
             playbackManager, &PlaybackManager::setAfterPlaybackAlways);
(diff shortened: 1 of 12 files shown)


tbertels:
I suggest to enable that playback time short mode setting by default.
I don't think many people need so the milliseconds in normal usage. It reduces the usability.

cmdrkotori:
>I suggest to enable that playback time short mode setting by default.
>I don't think many people need so the milliseconds in normal usage. It reduces the usability.
Yes that's probably a good idea as it will imitate MPC-HC out of the box. msec is for getting exact timestamps for making re-encoded clips (webms and the like), no reason to be so precise otherwise.

PR #169, opened 2024-09-11
Remember audio, video and subtitle tracks

tbertels:
- Save current video, audio and subtitle tracks for the current file.
- Restore current tracks on file load:
When a track is set to -1, it means the user didn't select it manually.
In that case, we check if an other track isn't better given the current settings.
Refactoring of the tracks selection.
Using the track id instead of the track name + id string simplifies the process.
- Reset the tracks menus actions groups on track change
This avoids accessing non existing actions in audioTrackSet, videoTrackSet and subtitleTrackSet.
- Includes a "temporary" fix to avoid setting subtitles to track 0 if using an older version of recents or favorites.

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -886,7 +886,8 @@ QString DisplayParser::parseMetadata(QVariantMap metaData,
 
 
 
-TrackInfo::TrackInfo(const QUrl &url, const QUuid &list, const QUuid &item, QString text, double length, double position)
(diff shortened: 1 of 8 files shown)


cmdrkotori:
Right, I accidentally merged your PRs in the reverse order. Can you please rebase this on master.

tbertels:
I don't think it would have mattered anyway, because I have to base commits from each pull request on master to prevent GitHub from showing changes from previous commits.

PR #175, opened 2024-09-13
Support adjusting subtitles delay

### LINKED ISSUE #94: Implement subtitle delaying
First, I want to let everyone that is working on this know that this is a great media player, and it is so close to being an equal replacement for MPC-HC. There are a few things either missing, or I can't find easily that would make it good enough to use daily. Sadly, I'm just a user so I don't know enough to help.
There seems to be no way to do subtitle delaying. I'm not sure what 'Increment' and 'Decrement' does. Maybe that speeds up the video playback? I'm unsure if there's a way to go to the next file in the list without going through the current video's chapters. Finally, there's no larger skip forward and backward only regular and finely. Hopefully some of this will come to this nice program.

tbertels:
- Support adjusting subtitles delay
By default F1 and F2 adjust the subtitles delay in a default step of 500 ms.
- Add a setting to change the default subtitles delay step
- Show the current subtitles delay in OSD when changing it
Shown as "Subtitles delay: xxx ms"

Fixes #94

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -613,6 +613,8 @@ void Flow::setupSettingsConnections()
             mainWindow, &MainWindow::setFullscreenHidePanels);
     connect(settingsWindow, &SettingsWindow::volumeMax,
             mainWindow, &MainWindow::setVolumeMax);
+    connect(settingsWindow, &SettingsWindow::subtitlesDelayStep,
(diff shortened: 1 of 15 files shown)


tbertels:
Don't merge yet, I'm looking into adding on change a message to the OSD with the current subtitles delay.

PR #177, opened 2024-09-15
Reduce logging time precision to 1 millisecond

tbertels:
Showing nanoseconds doesn't really make sense and just makes it harder to spot differences or copy/paste them.


cmdrkotori:
Good idea. Tbh I was blatantly copying the logging appearance of the linux kernel a bit too much.

PR #179, opened 2024-09-15
Fix typo in ActionEditor, add Indonesian and update translations

tbertels:
- Fix typo in ActionEditor
Mouse Fullscr -> Mouse Fullscreen
- Add Indonesian to translations
- Update translations strings
Indonesian was little more unsynced since it wasn't in the build


cmdrkotori:
Right, thanks for that I had completely forgotten that's how Qt works.

Issue #55, opened 2022-12-21
Pan and scan


flandriai:
It would be great if the **free stretch** option (that is present both in MPC-HC and MPC-BE on Windows) would be available in MPC-QT too. Currently this option is not available in any Linux media players, unfortunately. This gives you the possibility to stretch a video as you like both horizontally and vertically in realtime. _Not _zooming__, but **stretching**.  On Windows you can do this easily with the above mentioned players by using the numpad. (With MPC-HC/BE the image can also be moved freely within the window frame with the numpad keys, so the stretching can be extremely precise, this can be added to MPC-QT too.) It's very useful with a content that uses a non-standard aspect ratio (so black bars can be removed easily).

flandriai:
Updated the issue (deleted the part that is already added to the program and separated the two different issues which are addressing different areas. I opened a new issue for extrastreo feature.

tbertels:
Doesn't #185 fix this already now that extrastereo is in a different issue?
By default keys 4 and 6 change the aspect ratio (stretches the video in different directions) and key 9 disables it.
With MPC-HC/BE, the whole screen isn't always used, but how is that useful?
I'm not sure what's your use case here.

flandriai:
About extrastereo: you are right, it is separated by #185 but as a separate issue it can be marked as closed, when it is added to the program, and not cause any confusion (as if it has something to do with free stretch).

As for the free stretch option: In MPC-HC/BE there is also an option called as _"stretch to the window"_ (aside from disabling the aspect ratio that is already added to MPC-QT too.)
This makes the player to fill the entire window with the content horizontally. 
What useful free stretch option adds to this? Currently MPC-HC/BE is the only media player (as far as I know) with which you can resize a video in realtime without having to do any [re]encoding. Combined with the position change option (MPC-BE/HC is also able to move the content within the window both horizontally and vertically with assigned CTRL+ Numpad keys) this option makes it possible to watch any content conveniently without the possible black bars. Be it the most cutting edge monitor or an old CRT, every content can fill the whole screen. And it is not only useful with old content that uses 4:3 ratio, but also with  ones that does not use any standard aspect ratios, dimensions.
Also old contents frequently has "issues" on the top or on the very bottom, that can be cut down by this way too. The option is very precise in MPC-HC/BE, for stretching it uses a 0.01 value for positioning a 0.05 value.
[...]

PR #185, opened 2024-09-19
Support changing the video aspect ratio

### LINKED ISSUE #62: Stretch video to full screen.
There is no way to stretch 21:9 video to full screen in 16:9 format

tbertels:
Unlike MPC-HC, it's not a horizontal or vertical zooming/unzooming action, the video always uses as much space as possible given the aspect ratio.
An other advantage is that it requires only two keys instead of four.

#55
Fixes #62


tbertels:
These commands could be but into their own submenu inside Video, but I'm not sure if that's a good idea right now, with just these commands and the video streams.

PR #191, opened 2024-09-21
Add an option to show OSD timer on seek

tbertels:
- Add an option to show OSD timer on seek
In Tweaks, disabled by default.
- Reorder Tweaks section of settings (code style)

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1217,6 +1217,18 @@ void MainWindow::updateDiscList()
     ui->menuFileOpenDisc->setEnabled(addedSomething);
 }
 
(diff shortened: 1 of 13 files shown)


mainwindow.cpp
```diff
@@ -1817,6 +1829,11 @@ void MainWindow::setTimeTooltip(bool shown, bool above)
     timeTooltipAbove = above;
 }
 
+void MainWindow::SetOsdTimerOnSeek(bool enabled)
```
cmdrkotori, line 1832: camelCase

tbertels: My bad.

PR #192, opened 2024-09-22
Show aspect ratio name in OSD on change

tbertels:
Uses the aspect name, for example "70mm film".
Doesn't show anything if the aspect ratio doesn't have a name.


cmdrkotori:
Neat.  Something to think about for the future is if it should be attached to a checkbox in case someone doesn't wish to see these messages.

tbertels:
I'm guessing those who don't want OSD messages would just disable them completely, as its usefulness is more obvious and it's less obtrusive than OSD timer on seek.

PR #193, opened 2024-09-23
Don't show aspect ratio OSD message when opening a new file

tbertels:
- Don't convert eof-reached to bool
eof-reached is the first watched property notification that a new file is being opened.
On file change, eof-reached, which is then an empty string, can't be converted to bool, so it's set to default (false). Thus we can't know if it's a new file or a file being played again after reaching the end.
- Don't show aspect ratio OSD message when opening a new file
With the first file opened with mpc-qt, aspect-name isn't first set to empty, so OnFirstPlay gets skipped.
- Rename fileOpenedOrClosed to fileClosed now that it's more precise

### DIFF
--- a/manager.h
+++ b/manager.h
@@ -54,7 +54,7 @@ signals:
     void videoSizeChanged(QSize size);
     void playbackSpeedChanged(double speed);
     void stateChanged(PlaybackManager::PlaybackState state);
-    void fileOpenedOrClosed();
+    void fileClosed();
     void typeChanged(PlaybackManager::PlaybackType type);
     // Transmit a map of chapter index to time,description pairs
     void chaptersAvailable(QList<QPair<double,QString>> chapters);
--- a/manager.h
+++ b/manager.h
@@ -179,7 +179,7 @@ private slots:
     void mpvw_pausedChanged(bool yes);
     void mpvw_playbackIdling();
     void mpvw_playbackFinished();
-    void mpvw_eofReachedChanged(bool eof);
+    void mpvw_eofReachedChanged(QString eof);
     void mpvw_mediaTitleChanged(QString title);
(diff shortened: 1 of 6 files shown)


manager.h
```diff
@@ -4,6 +4,8 @@
 // widget.  Due to needing to communicate with the playlists, it sort of takes
 // control of the mpv widget away from its host.
 
+#define STR_TRUE "true"
+#define STR_FALSE "false"
```
cmdrkotori, line 8: Make these static const char variables in manager.cpp, such as strTrue and strFalse.  See main.cpp for example.

manager.h
```diff
@@ -157,6 +159,7 @@ public slots:
                              int64_t &audioTrack, int64_t &subtitleTrack);
 
 private:
+    enum aspectNameChanged { OnOpen, OnFirstPlay, Manually };
```
cmdrkotori, line 162: enums are CapitalCase

manager.h
cmdrkotori, line 8: Make these static const char[] in manager.cpp.  For example the top main.cpp.

manager.h
```diff
@@ -240,6 +243,8 @@ private slots:
 
     bool timeShortMode = false;
 
+    PlaybackManager::aspectNameChanged showAspectOsdTriggeredBy = aspectNameChanged::OnOpen;
```
cmdrkotori, line 246: aspectNameChanged -> AspectNameChanged

cmdrkotori:
Apologies for the notification mess, it seems I pressed the wrong button.

tbertels:
> Apologies for the notification mess, it seems I pressed the wrong button.

No problem. 😉

PR #196, opened 2024-10-05
Replace Fine (seek) step by Large step, Time step by Normal step

### LINKED ISSUE #117: Feature Request: Custom jump distances
Add the ability to add custom jump distance
image

tbertels:
- Replace Fine (seek) step by Large step, Time step by Normal step
Users probably need a large seek step more than a finer seek step (than normal).
If users really want a finer seek step and we can find a way to fit a third option, we could add it later.
Fixes #117
- Reorder seek commands in options and frame step commands
Reorder actions so that seek actions aren't at the bottom in keys
settings.
Reorder seek and frame step commands so that back comes before next
(like for most other commands).
- Reorder Previous and Next subtitle commands
Reorder previous and next subtitle commands so that their order matches
the order of all other commands (previous at the top, next at the
bottom).

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -640,10 +640,10 @@ void Flow::setupSettingsConnections()
             playbackManager, &PlaybackManager::setSpeedStep);
     connect(settingsWindow, &SettingsWindow::speedStepAdditive,
             playbackManager, &PlaybackManager::setSpeedStepAdditive);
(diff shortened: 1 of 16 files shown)


cmdrkotori:
LGTM.  Some users may need to reset their shortcut keys but that's what happens when you migrate between versions.

tbertels:
It may be a good idea to mention in the changelog that this is a breaking change.

Issue #200, opened 2024-10-10
VolumeSlider widget often doesn't have a background


tbertels:
Here's what it looks like:

![Copie d'écran_20241010_084219](https://github.com/user-attachments/assets/ebbd7a6b-a7b6-4927-bd8b-c095754f8381)

Or sometimes

![Copie d'écran_20241010_084341](https://github.com/user-attachments/assets/f9ef5475-8049-4680-80b0-cb1a83d819f0)

Instead of

![image](https://github.com/user-attachments/assets/4157860f-4cc7-48ee-8958-0e1f7b674bd9)

I've bisected the issue to https://github.com/mpc-qt/mpc-qt/commit/be267c475ebe8f41c6e0e6414cdaa5399c521b24

The fact that it's not always reproduceable (only 9 times out of 10) and of course the sometimes partial (and with wrong polygon shape) background makes me think that it's a rendering bug, so maybe a Qt bug?

I've reproduced it with X11, Xwayland and Wayland.

tbertels:
Here's what I see in the log when always redrawing the background (https://github.com/mpc-qt/mpc-qt/pull/201):

> [7.754] [drawnslider] MediaSlider::makeBackground
[7.759] [drawnslider] VolumeSlider::makeBackground
[7.759] [qt] warn: QPainter::begin: Paint device returned engine == 0, type: 3
[7.759] [qt] warn: QPainter::scale: Painter not active
[7.759] [qt] warn: QPainter::setRenderHint: Painter must be active to set rendering hints
[7.759] [qt] warn: QPainter::setPen: Painter not active
[7.759] [qt] warn: QPainter::setBrush: Painter not active

cmdrkotori:
Yeah, paddingHeight was passed through the constructor, but not saved to member variable.  I'll push a fix to master.

PR #201, opened 2024-10-10
Always redraw the seekbar

tbertels:
This prevents chapters marks from sometimes disappearing on start after being shown.


tbertels:
This doesn't fix  #200 though.

PR #205, opened 2024-10-10
Default options

tbertels:
- Don't use additive speed step by default
Additive makes it harder to reset to normal speed: decreasing then increasing the speed doesn't make it come back to 100%.
- Limit the volume to 100% by default
Allows users to easily change the volume without risking clipping.

### DIFF
--- a/manager.h
+++ b/manager.h
@@ -209,7 +209,7 @@ private:
     double mpvLength = 0.0;
     double mpvSpeed = 1.0;
     double speedStep = 2.0;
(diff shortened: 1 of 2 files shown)


tbertels:
About the volume boost, mpc-hc has one in Audio switcher: up to +300%.
mpv allows volume max up to 1000 instead of 100.
So an option for that would probably be better in Audio instead of a checkbox in Tweaks.

cmdrkotori:
Yes that would be a better place for it.

PR #207, opened 2024-10-11
Audio settings

tbertels:
- Implement audio balance setting
- Add live preview for audio balance setting
- Add tooltip to volume setting


tbertels:
Here's what it looks like with the Plasma Breeze theme as a QScrollBar:

image

It can be changed to QSlider but then there's a color that shows the "increase":

image

It can be hidden with a stylesheet, but then it won't follow the theme (except for dark/light).
Here are examples:

image

image

cmdrkotori:
The logical conclusion here is a custom widget. In a pinch we could subclass the MediaSlider/DrawnSlider widget, with custom ticks at every 10%, perhaps with extra code to snap the selected value to the nearest 1%.

But for now a scrollbar will do.  I tested this and it works a treat.

PR #209, opened 2024-10-12
Simplify the "After Playback" actions

tbertels:
Group all actions in one action group, remove the distinction between "Once" and "Always".
This distinction is more likely to confuse users into thinking that "Always" means the setting is saved accross restarts.
Repeat (Once): not very useful, it just repeats the file once.
Exit (Always): duplicate of Exit (Once) because the action isn't saved.
Do Nothing (Always): not needed anymore with just one action group.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -535,13 +535,6 @@ void MainWindow::setupMenu()
     ui->actionViewHideCapture->setVisible(false);
     ui->actionViewHideNavigation->setVisible(false);
 
-    // Work around separators with text in the designer not showing as
-    // sections
-    ui->menuPlayAfter->insertSection(ui->actionPlayAfterOnceExit,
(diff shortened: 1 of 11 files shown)


tbertels:
I think the "Repeat" After Playback action should replaced by the Repeat action in Loop, with renaming Loop as Repeat.
The "Repeat" After Playback action is very slow compared to the Repeat action in Loop.

cmdrkotori:
>I think the "Repeat" After Playback action should be replaced by the Repeat action in Loop, with renaming Loop as Repeat.
Yeah and "Repeat" in the "After Playback" location isn't in MPC-BE or MPC-HC.  It's useless clutter that could be removed now that looping is there.

>The "Repeat" After Playback action is very slow compared to the Repeat action in Loop.
Yep it re-opens the file instead of looping what's already loaded.

PR #219, opened 2024-10-17
Restrict next/previous file feature to audio/video files

### LINKED ISSUE #158: "Skip Backward/Forward" Opens Non-Media Files
When using the "Skip Forward" feature in MPC-QT, instead of moving to the next playable media file in the folder (such as music or video files), the player opens all types of files indiscriminately, including non-media files like PDFs, TXT files, and other unsupported file types. This behavior differs from what is seen in other players like MPC-HC or MPC-BE, where only valid media files (e.g., MP4, MKV, MP3) are opened.

### Steps to reproduce:

Open a media file in MPC-QT from a folder containing various file types (including PDFs, text files, and media files).
Press the "Skip Forward" button.
Observe that the next file is opened, but it may be a non-media file (e.g., a PDF or TXT file).

**Expected behavior**: The "Skip Forward" feature should only open playable media files, such as music or video files (e.g., MP3, MP4, MKV), similar to the behavior of MPC-HC or MPC-BE.

**Actual behavior**: "Skip Forward" opens all file types, including non-media files like PDFs and TXT files, leading to an undesirable user experience when navigating through a folder of mixed file types.

### Proposed Solution:

Add functionality to filter files based on their MIME type or file extension, so that only media files are recognized and opened during navigation.
 Alternatively, implement a file filter setting where users can specify what types of files should be considered for playback (e.g., MP3, MP4, MKV).

### Environment:

**MPC-QT version**: I compiled the latest git version "commit 78fd54d"
**Operating System**: arch Linux i use KDE + wayland 
**Additional Information:**
    
- There is an issue where the same cover art is displayed for every track, even when different files have different cover art.
- Due to this bug, MPC-QT cannot be resized to a smaller window, as the large cover art makes it difficult to reduce the size.
- It would be helpful to add an option to disable cover art entirely, so users can use MPC-QT as a music player without cover art display, allowing for a more compact interface.

tbertels:
- Restrict next/previous file feature to audio/video files
This prevents archives or pictures in the folder from getting opened. This still allows manually opening non audio/video files or playing them if they're in a playlist.
- Cleanup file extension before filtering
This makes it possible to play incorrectly named files with a name like "video.mkv (1)".

Fixes #158

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -17,7 +17,7 @@
 #include "helpers.h"
 #include "platform/unify.h"
 
-QSet<QString> Helpers::fileExtensions {
+QSet<QString> Helpers::audioVideoFileExtensions {
     // DVD/Blu-ray audio formats
     "ac3", "a52",
     "eac3",
@@ -118,6 +118,11 @@ QSet<QString> Helpers::fileExtensions {
     "m3u", "m3u8",
     "pls",
     "cue",
+    // Modfiles
+    "mod"
+};
+
+QSet<QString> Helpers::imagesFileExtensions {
     // Image formats
     "bmp",
     "dds",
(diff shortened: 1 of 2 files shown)


tbertels:
We should use QMimeType in the future so that even files with the wrong extension are correctly identified.

cmdrkotori:
>We should use QMimeDatabase in the future so that even files with the wrong extension are correctly identified.

Indeed.  The [mpv desktop file](https://github.com/mpv-player/mpv/blob/master/etc/mpv.desktop) has a list of playable mime types that usually get associated with the mpv player.  I don't think it's exhaustive though because it's lacking mime types for supported images.

Issue #21, opened 2022-02-15
Feature Requests


Dejv311:
Hi,
just discovered mpc-qt and it's pretty awesome! Some features I am missing:

The popup messages currently show only when increasing/decreasing playback speed and they are huge
The buttons under the seekbar had little gray dividers in the original which made it much easier to "read"
For some reason up/down is not mapped to volume up/down and I couldn't even find the option to map it so
Drag&Drop does not work for me on Ubuntu 20.04, which is kinda a bummer :/
Ctrl+A (aka always on top mode) does not switch back to "default" when pressed again like in the original

That's it for now :) Many thanks for making this marvel!

AlexFolland:
Disclaimer: I am another user, not a developer of this project.  I am posting this commentary to help clarify this issue ticket.

> Some features I am missing:

This looks like a list of issues when comparing to some other program, not missing features.  What program?

> The popup messages currently show only when increasing/decreasing playback speed and they are huge

OSD messages appear when changing volume here and look good; can't reproduce.

![image](https://user-images.githubusercontent.com/601517/154638314-91268a3a-690e-4e0e-9405-2c6892022ccf.png)

> The buttons under the seekbar had little gray dividers in the original which made it much easier to "read"

What's "the original"?  Can you please paste a screenshot to show what you're talking about?

> For some reason up/down is not mapped to volume up/down and I couldn't even find the option to map it so

These binds are in the "Keys" menu, but labelled "Up" and "Down".  IMO they should be named "Volume Up" and "Volume Down" instead, for clarity, which would resolve this issue.  Here's a screenshot of them.  You can configure them here.

![image](https://user-images.githubusercontent.com/601517/154638582-dd6cfded-c13e-476f-9235-e036c6aebcda.png)

> Drag&Drop does not work for me on Ubuntu 20.04, which is kinda a bummer :/

What are you dragging and dropping, and from where?  I can drag subtitles onto the video and they are immediately loaded here on Manjaro XFCE with the `mpc-hc-git` AUR package.  I can also drag entries into the playlist and drag videos directly onto the viewport, which overwrites the playlist.  This is with the latest `xfwm4-git` AUR build as of this comment and I'm dragging files from the Nemo file manager.  Are you using Wayland?

> Ctrl+A (aka always on top mode) does not switch back to "default" when pressed again like in the original

[...]

PR #224, opened 2024-10-18
Implement OSD font and font size settings

tbertels:
#21

### DIFF
--- a/settingswindow.cpp
+++ b/settingswindow.cpp
@@ -499,9 +499,6 @@ void SettingsWindow::setupUnimplementedWidgets()
     // Remove the trailing : (looks odd with the rest of the tooltip options hidden)
     ui->tweaksTimeTooltip->setText(ui->tweaksTimeTooltip->text().replace(":",""));
     ui->tweaksTimeTooltipLocation->setVisible(false);
-    ui->tweaksOsdFont->setVisible(false);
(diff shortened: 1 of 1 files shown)


tbertels:
It would look better if the settings were aligned on the left or stretched to take the full width though.

cmdrkotori:
>It would look better if the settings were aligned on the left or stretched to take the full width though.

Yep, alignment is a constant headache when designing gui. At the moment it's aligning with the hidden combo box of time tooltip location.  If we delete that combo box and drag the size of "Use time tooltip" out to the entire width, the osd widgets will align as expected.

tbertels:
> > It would look better if the settings were aligned on the left or stretched to take the full width though.
> 
> Yep, alignment is a constant headache when designing gui. At the moment it's aligning with the hidden combo box of time tooltip location. If we delete that combo box and drag the size of "Use time tooltip" out to the entire width, the osd widgets will align as expected.

Indeed!
I ended up unhiding time tooltip, adding a checkbox for OSD font and changing the wording. It's not that great but better, and only with English locale.
It would probably be better to use a stylesheet for the whole app and show it in the Stylesheet setting so that users can easily customize it. Because a stylesheet makes it possible to change the alignment, which should fix this problem.

PR #180, opened 2024-09-15
Native file dialog of Qt6 will crash app on macOS Ventura(Maybe sonoma too).

face1es5:
Use qt-native file dialog if current system is macOS.

PR #227, opened 2024-10-20
Remove remaining mac code

tbertels:


cmdrkotori:
This reverts #180.  I'm not against it.  Apple should fix their bugs rather than having people like us patch around them.

tbertels:
> This reverts #180. I'm not against it. Apple should fix their bugs rather than having people like us patch around them.

I thought this was remaining old code from before the mac code removal.
See #187 and efa5c64dcb49cb2b49ef11cbc717bc85f2ed03d2.

If this is actually useful, I think it can be kept.

PR #237, opened 2024-10-25
Increase recent files size to 1000 when remember position is enabled

tbertels:
But only show the 20 most recent ones in the menu.

I've tested it on a 2013 Intel Core i3-4130 @ 3.40GHz with an SSD, where it takes about 10 ms more to load the recent.json file.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -395,7 +395,9 @@ void Flow::readConfig()
         QVariantMap favoriteMap = storage.readVMap(fileFavorites);
         favoriteFiles = TrackInfo::tracksFromVList(favoriteMap.value(keyFiles).toList());
         favoriteStreams = TrackInfo::tracksFromVList(favoriteMap.value(keyStreams).toList());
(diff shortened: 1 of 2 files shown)


tbertels:
Alternatively, we could add a Tweaks setting to increase the recent files to 1000.

cmdrkotori:
This is a good idea, we need a way to stress-test our algorithms.

tbertels:
When you say it's a good idea, what are you talking about?

cmdrkotori:
What I mean is that the speed of doing things will become evident with larger data files. If we test with 1000 entries of json in the recents file it will be more obvious if we happen to introduce polynomial behaviour.

tbertels:
All right, I wasn't sure whether you were also talking about the Tweaks setting.

On a different note, as far as I'm concerned, we could create the release branch right now.

cmdrkotori:
Done.

https://github.com/mpc-qt/mpc-qt/tree/release/24.12

PR #238, opened 2024-10-29
Don't repeat file when playlist in shuffle mode with only one file

tbertels:
The shuffle mode currently picks randomly the next file just before playing it, so if there's only one file, it'll play it in loop instead of stopping.

In the future, we should probably randomize the whole playlist (but keep the sort order) and save that state across restarts. That way, the same file won't play twice, it'll be possible to go back to the previously played track and the playlist will have an actual end.

### DIFF
--- a/playlistwindow.cpp
+++ b/playlistwindow.cpp
@@ -122,7 +122,7 @@ PlaylistItem PlaylistWindow::getItemAfter(QUuid list, QUuid item)
     } else {
         after = pl->itemAfter(item);
     }
-    if (!after)
+    if (!after || after->uuid() == item)
         return { QUuid(), QUuid() };
     return { pl->uuid(), after->uuid() };
 }


tbertels:
We may want to pick this for release, though it's not critically needed.

cmdrkotori:
>we should probably randomize the whole playlist (but keep the sort order) and save that state across restarts.

Good idea.

I was thinking about implementing this once and I had landed on adding a `nextToPlay` or `queue` list to the `Playlist` class.  In normal mode it would have the same order as items, but in shuffle mode it would be random.  I think it's still there in a comment.

The naive way to do this is to augment the insertion/deletion functions to do random insertions/deletions on the queue in addition to the existing code.  The problem happens when a lot of items in the playlist are deleted, searching for the corresponding item to delete in the nextToPlay queue is an O(n) operation, and doing that O(m) times (however long the selection is) results in an O(m\*n) (quadratic) operation.  Likewise, *m* random insertions in an *n*-size list is O(m\*n).

The solution of course is to toss out the old shuffle order completely whenever an insert/delete happens and remake the shuffle order when operations are finished.  The benefit of this approach is that there is no need need to maintain the nextToPlay queue inside `Playlist::addItem` `Playlist::removeItem` et al, we can simply invoke reshuffle at the end of every appropriate function in PlaylistWindow.

So while I would like to do the naive way because it will always preserve order, I don't think it can be helped because there will be people out there with 100,000 items (their entire music library for example) in the playlist.

tbertels:
> The naive way to do this is to augment the insertion/deletion functions to do random insertions/deletions on the queue in addition to the existing code. The problem happens when a lot of items in the playlist are deleted, searching for the corresponding item to delete in the nextToPlay queue is an O(n) operation, and doing that O(m) times (however long the selection is) results in an O(m*n) (quadratic) operation. Likewise, _m_ random insertions in an _n_-size list is O(m*n).

It could probably be made faster than reshuffling the whole list by only iterating through the list once, and either:
- comparing each item with the list of items to delete
- inserting each item in a predetermined position once we get there

tbertels:
After experimenting a bit with two QHash with chained uuids for Playlist::itemAfter and Playlist::itemBefore, I'm wondering if the best solution wouldn't be to show the shuffled order in the playlist, like the Randomize feature does.

So in shuffle mode, the list would get randomized by the Randomize feature, with original position already saved by that feature.
The original position would be saved in the playlists file if we're in shuffle mode so that the original position can be restored when the user disables Shuffle mode.
The Randomize menu action would be renamed "Reshuffle" and would do the same thing as disabling then re-enabling Shuffle mode.

This would have several advantages:
- Simpler code as existing functions to play next/previous or insert, move and remove items can be reused
- Fast code since the list would work the same way as right now
- Minimal overhead in the playlists file (just one originalposition int property per file)
- The user would be able to see the playing order of files and how many files are remaining

About the insert function, I'm not sure it's a good idea to do random inserts. How can the user know that the files got correctly inserted if they are spread in the list?
I think it would be better to let the user reshuffle the list if needed.

cmdrkotori:
That sounds good. :+1:

cmdrkotori:
7a021293f736c4e691b4856fa7dcf0c9b00ef6c6 has been cherry-picked on to release/24.12

PR #241, opened 2024-11-02
Full playlist shuffle mode

tbertels:
- Implement full playlist shuffle mode
When shuffle mode is enabled, the original positions of the playlist items are saved, and the playlist is visibly shuffled. Each item is only played once and the next item is the item below the current item.
The currently playing item is kept as the first item in shuffle mode.
Once shuffle mode is disabled, the items are restored to their original positions.
The Randomize feature is replaced by a Reshuffle feature in the shuffle mode.
The Sort and Restore features are disabled in shuffle mode.
When the user adds items, they are added to the end of the playlist.
- Add logging for playlist shuffle and files load
DrawnPlaylist::repopulateItems() (called by PlaylistWindow::refreshPlaylist) is by far the slowest step of shuffling the playlist.
On a 2013 Intel Core i3-4130 @ 3.40GHz with a SATA SSD and a 200 000 items playlist (which takes up 62 MB on disk):
- - shufflePlaylist takes 15 ms
- - refreshPlaylist takes 461 ms
Reading that playlist on start takes 812 ms.
- Reshuffle the playlist on repeat
Once we reach the end of the playlist in repeat mode, reshuffle it before restarting playing it.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -395,9 +395,7 @@ void Flow::readConfig()
         QVariantMap favoriteMap = storage.readVMap(fileFavorites);
         favoriteFiles = TrackInfo::tracksFromVList(favoriteMap.value(keyFiles).toList());
         favoriteStreams = TrackInfo::tracksFromVList(favoriteMap.value(keyStreams).toList());
-        LogStream("main") << "reading fileRecent start";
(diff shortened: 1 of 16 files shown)


tbertels:
#238 has more context.

PR #244, opened 2024-11-03
Add the default playlist format string in editable form

tbertels:
This makes it easier to edit for the user than just a placeholder text.

### DIFF
--- a/settingswindow.ui
+++ b/settingswindow.ui
@@ -5645,7 +5645,7 @@ media file played</string>
               <item>
                <widget class="QLineEdit" name="playlistFormat">
                 <property name="text">
-                 <string/>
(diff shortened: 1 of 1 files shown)


tbertels:
The default playlist format includes the track number (as in CD track), but this is only shown once the file is playing, so it's not very useful even for full albums.
This results in variable display for example "1. Artist - Title" or "3/3. Artist - Title" depending on the file when multiple files from multiple albums are mixed.
Does it benefit most users to show this by default?
I haven't seen other media or music players doing this, so I'd suggest removing it by default.

cmdrkotori:
You're right, it doesn't make much sense on big playlists.

PR #246, opened 2024-11-04
Don't bail out until lock file can be locked or we get a response

tbertels:
Since 1f88fb3507ee5b18b5719614db33db258b454a40 if the user tries to open mpc-qt while the previous instance is still closing (usually because it's writing the playlists file), the new instance closes immediately.

This ensures it will wait until the lock file can be locked or until an other main instance replies.


tbertels:
We probably want to backport this to [release](https://github.com/mpc-qt/mpc-qt/tree/release/24.12).

cmdrkotori:
7d5f30e has been backported to release/24.12

PR #247, opened 2024-11-05
Send the request to main server while waiting for lock file

tbertels:
7d5f30edd53747327948ffdf79e901ef41ae7e04 intended to do this, it was a mistake.

Without this, we're kinda back to what it was before 1f88fb3507ee5b18b5719614db33db258b454a40.
The difference is that we get zombie instance(s) that become alive in turn once the main instance gets closed.


cmdrkotori:
ea15e064a2bfbded53be4ef815362e47816b7b14 was backported to release/24.12

PR #248, opened 2024-11-06
Implement Go To feature

tbertels:
If the regex validator is set to prevent entering incorrect times (like 63), it doesn't play nice with the event filter.
I first tried with an input mask, but it's doesn't look nice and above all it leads to ugly bugs, like incorrect seeking.

The frame number may not be fully accurate because the fps is the estimated-vf-fps. We may want to show the fps somewhere like mpc-hc does (it actually allows to change it too).


cmdrkotori:
Very neat, good job.

tbertels:
Thanks!

PR #255, opened 2024-11-11
Check if playlist pointer is valid before accessing shuffle property

tbertels:
This crash happens on exit when playlist backups don't already contain the repeat attribute.
This check means that until backups get updated (and thus contain repeat) by removing or restoring a playlist, they won't get written to disk.
Regression caused by efceb5e1da37550e51195fb74d2aea9de3557679.

### DIFF
--- a/playlist.cpp
+++ b/playlist.cpp
@@ -156,8 +156,9 @@ QVariantMap Item::toVMap() const
     v.insert(keyUrl, url());
     v.insert(keyUuid, uuid());
     v.insert(keyMetadata, metadata());
-    if (PlaylistCollection::getSingleton()->getPlaylist(playlistUuid())->shuffle())
-        v.insert(keyOriginalPosition, originalPosition());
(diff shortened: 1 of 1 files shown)


tbertels:
Note that I didn't read enough playlist code to understand why this does actually happen until the backups get updated (as opposed to never).

This also makes me think that we could only write backups to disk (in normal usage) if they've changed.

cmdrkotori:
>This also makes me think that we could only write backups to disk (in normal usage) if they've changed.

Probably a good idea.  The programs has thus far used write-at-exit to avoid disk thrashing.

tbertels:
What I meant is only write them **on exit** if they've changed.

cmdrkotori:
Right, well the quick fix for that is to modify Storage::readJsonObject and Storage::writeJsonObject to cache the read json data to a static member variable, and only write it out if it hasn't seen the data before (or if the file no longer exists).

PR #258, opened 2024-11-12
Fix slider position with Hi-DPI

### LINKED ISSUE #256: Timeline slider isn't vertically centered on timeline
Regression caused by be267c475ebe8f41c6e0e6414cdaa5399c521b24.

Before:

Copie d'écran_20241111_155718

Now:

Copie d'écran_20241110_135607

cmdrkotori:
Vertical position is fixed with Hi-DPI.

Fixes #256.

### DIFF
--- a/widgets/drawnslider.cpp
+++ b/widgets/drawnslider.cpp
@@ -173,9 +173,8 @@ void DrawnSlider::paintEvent(QPaintEvent *event)
         double px;
         double x = isDragging ? xPosition : valueToX(value());
         x -= handleWidth/2.0;
-        x *= pr;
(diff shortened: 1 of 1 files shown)


cmdrkotori:
Testing Hi-DPI has revealed a different visual bug: the slider appears to jump from pixel to pixel as the time progresses instead of subpixel progression.  But only with Hi-DPI; 100% works fine.

The last time I remember testing this was with Qt 5, and it worked then.  So I suspect something subtly changed that I'm not aware of.

tbertels:
Nice!
Obviously, we do want this for 24.12.

cmdrkotori:
58353bd has been backported to release/24.12.

PR #260, opened 2024-11-13
Don't write playlists backup to disk on exit if not changed and add logging for files write

tbertels:
- Don't write playlists backup to disk on exit if not changed
Based on idea of @cmdrkotori in #255.
- Add logging for files write

### DIFF
--- a/storage.h
+++ b/storage.h
@@ -1,7 +1,16 @@
 #ifndef STORAGE_H
 #define STORAGE_H
 
+#include <QJsonDocument>
 #include <QObject>
+
+extern const char fileFavorites[];
+extern const char fileGeometryV2[];
+extern const char fileKeys[];
+extern const char filePlaylists[];
+extern const char filePlaylistsBackup[];
+extern const char fileRecent[];
+extern const char fileSettings[];
 
 class Storage : public QObject
 {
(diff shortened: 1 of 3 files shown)


tbertels:
Note that because of a bug with playlist shuffling, the backup gets written every time if there's a playlist backup where shuffle is true and originalposition isn't set.

storage.h
```diff
@@ -1,8 +1,17 @@
...
+#include <QJsonDocument>
 #include <QObject>
 
+static const char fileFavorites[] = "favorites";
+static const char fileGeometryV2[] = "geometry_v2";
+static const char fileKeys[] = "keys_v2";
+static const char filePlaylists[] = "playlists";
+static const char filePlaylistsBackup[] = "playlists_backup";
+static const char fileRecent[] = "recent";
+static const char fileSettings[] = "settings";
```
cmdrkotori, line 13: Data shouldn't be in header files.  I suggest making these public static members of the Storage class, and placing their definition at the top of Storage.cpp.  main.cpp can be altered to introduce `Storage::` where it refers to these variables.

cmdrkotori (commented):
Looks good so far.

storage.h
tbertels: `extern` seems like a better choice.
Those in helpers.h should be changed too.

The other ones (used locally) can be replaced by C++11 `constexpr`.

PR #261, opened 2024-11-14
Use constexpr or extern for static const char

tbertels:
- Move definition of const char out of header file
See https://github.com/mpc-qt/mpc-qt/pull/260#discussion_r1841676590
- Use C++11 constexpr char instead of static const char

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -17,6 +17,9 @@
 #include "helpers.h"
 #include "platform/unify.h"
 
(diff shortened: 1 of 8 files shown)


cmdrkotori:
Thanks for this.  When I went through uni C++11 wasn't a thing.  I've tried my best to keep up with newer additions to C++ since then, but constexpr apparently was one of the things I missed.

tbertels:
You're welcome! 😉

PR #263, opened 2024-11-14
Handle SIGHUP, SIGINT and SIGTERM signals

tbertels:
This allows mpc-qt to write files to disk before closing when it receives these signals.

For reference:
SIGHUP: sent when the terminal is closed
SIGINT: sent when CTRL+C is pressed in the terminal
SIGTERM: sent by default by kill and killall commands


cmdrkotori:

PR #264, opened 2024-11-17
Ensure mpc-qt closes correctly when handling signals

tbertels:
- Use QCoreApplication::exit when handling IPC signals
When using quit, if the playlist is opened, mpc-qt tends to hang with the window hidden instead of closing.
- Move settings and recent update to the destructor
So that they're called when handling IPC signals.
- Add logging in the destructor


cmdrkotori:

PR #265, opened 2024-11-18
Add smoke tests CI workflow

tbertels:
- Add smoke tests CI workflow
This checks if mpc-qt starts and stops correctly.
It also checks if the various JSON files are written to disk.
Xvfb is used as there's no display in the container.
- Don't limit CI workflows to master branch
It'll be useful for the release branch.
- Improve CI workflow messages


tbertels:
We most likely want this for release.
Along with #259, #263 and #264.
I actually found out about those while doing these tests.

After backporting this PR, the tests should fail until those PRs are backported.

cmdrkotori:

PR #269, opened 2024-11-20
Use ldd for msys2 DLLs

tbertels:
This ensures we don't need to update the list anymore.
We still need to manually copy Qt6Svg.dll as it's dynamically loaded.
Idea from https://github.com/mpc-qt/mpc-qt/commit/4b73ee71b22d08cbf19a07e26b4b0dc69108893e.

### DIFF
--- a/make-release-msys2.sh
+++ b/make-release-msys2.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/bash
 # Like make-release-win.sh, but everything comes from the msys2 installation
 
+set -e
+
 VERSION=""
 if [ -v GITHUB_SHA ]; then
     VERSION=`date +'%Y.%m.%d'`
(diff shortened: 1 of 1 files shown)


cmdrkotori:
Definitely an improvement.  But it appears to be missing Qt6Svg.dll. The buttons are blank, but it works if you copy the dll over.

I think this dll is loaded dynamically and not statically.  So either we copy it manually, or we write a small snippet of dummy code somewhere that pulls QtSvg in.

tbertels:
Checking the DLLs, I've noticed that Qt6Xml.dll was manually copied but isn't detected either.
Do we actually use it?

cmdrkotori:
> Checking the DLLs, I've noticed that Qt6Xml.dll was manually copied but isn't detected either. Do we actually use it?

Apparently not, because the program runs fine.

tbertels:

PR #271, opened 2024-11-21
Check if tracks menu action exists before trying to access it

tbertels:
With:
- file_1 which has two audio tracks and track_1 selected
- file_2 which has three audio tracks and track_3 selected
If we open file_1 then file_2, audioTrackSet gets called to mark track_3 as selected before the audio tracks menu has been updated to contain three tracks instead of two, thus segfaulting.

This doesn't seem to happen with subtitles and I haven't tested it with video tracks, but better be safe I guess.


tbertels:
Good candidate for backporting to release.

PR #270, opened 2024-11-20
Add version information to log

tbertels:
This may make troubleshooting easier.

PR #272, opened 2024-11-22
Release2412 update 2

cmdrkotori:
Pulls #269.  Alleviate the need to keep the release branch's build processes updated.
Pulls #270.  I'd rather we have this for debugging future issues after release.
Pulls #271.  Avoid segfault in the track selection.
Pulls #273.  Build MSYS2 artifacts inside release branches.

### DIFF
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -2,9 +2,7 @@ name: Msys2 CI
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


cmdrkotori:
If you think this is too much, let me know and I'll remove some of these.

tbertels:
No, it's exactly what I had in mind.

PR #276, opened 2024-11-25
Logging improvements

cmdrkotori:
Turn the mechanism where logs can be pushed on to stderr into a full-blown feature.  This new feature is invoked with the argument `--log-to-console`.  This is principally a debugging feature.

**Why shouldn't I do this?** The log thread will flush stderr after every line, which may slow it down.


tbertels:
Very nice!

I suggest rewording "Block the main thread while log flush completed" to "Block the main thread while log flush completes".

cmdrkotori:
Done.

PR #278, opened 2024-11-26
Flush pending messages to disk on fatal message

cmdrkotori:
Also flush any pending messages to the log file.  Most users on Windows would not see the pending message dump on stderr, and depend on the file to retrieve the log.  This may make fatal errors reported by qFatal more complete.  However, if the logs are not set to immediate mode and a segfault happens between dumps to disk, messages will still get missed.

Change "Window update interval" to "Update interval" to indicate that it applies across the board.  Change update interval suffix from "msec" to "ms".  Indicate that "No interval" should be used during testing.

### DIFF
--- a/logger.cpp
+++ b/logger.cpp
@@ -42,9 +42,9 @@ void loggerCallback(QtMsgType type, const QMessageLogContext &context, const QSt
         Logger::log("qt", "crit", msg);
         break;
     case QtFatalMsg:
-        Logger::fatalMessage();
-        fprintf(realStdErr.ptr, "[FATALITY] [qt] fatal: %s\n", msg.toUtf8().data());
-        fflush(realStdErr.ptr);
+        Logger::log("qt", "fatal", msg);
+        QMetaObject::invokeMethod(Logger::singleton(), "fatalMessage",
+                                  Qt::BlockingQueuedConnection);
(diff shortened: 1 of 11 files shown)


cmdrkotori:
I guess if we're feeling brave we can intercept the SIGSEGV signal during the lifetime of the logger and use it flush pending logging messages in the case that immediate mode is not selected.

tbertels:
While you're at it, you could rename the `logUpdateInterval` suffix from msec to ms.

cmdrkotori:
Fixed. `fatalMessage()` is now a blocking call to the log thread, the race condition should be no more.

> While you're at it, you could rename the `logUpdateInterval` suffix from msec to ms.

Done.

cmdrkotori:
Ok that should be it for real this time.  I thought of a last-minute bug an hour after I've pushed. (the need to empty the pending message queue). Sorry for the mail spam.

tbertels:
It works fine with qFatal. 👍

cmdrkotori:
>Do you mean that there would be side effects?

>I've written some code to do that and it doesn't seem to affect the stack trace.

Well that's good. I thought we might have to think about using libbacktrace to create the traces ourselves.

PR #295, opened 2024-12-05
Work around KDE bug to hide cursor on rightmost pixels in fullscreen

### LINKED ISSUE #294: Mouse cursor not hidden on rightmost pixels (Plasma 6.2.4)
With Plasma 6.2.4 and X11 (works fine with Wayland) and display scaling set to anything else than 100%, the mouse cursor stays visible on rightmost pixels of the screen or window both in windowed and fullscreen mode.
Tested with default mpc-qt config.

I've checked with Plasma 6.1.5, it works fine.

I can't reproduce it in a VM.
@cmdrkotori Can you reproduce it on your system (not a VM) with X11?

The widget is 1 pixel wider than the window.
Shown by adding this to `MpvGlWidget::paintGL()`:
```
QRect widgetRect = QRect(mapToGlobal(QPoint(0,0)), size());
QRect windowRect = window()->frameGeometry();
LogStream("glwidget") << "Widget rect: " << widgetRect.right() << " Window rect: " << windowRect.right();
```

The only way I've found to fix it so far is to hide the cursor for the whole window in `MpvObject::hideCursor()`:
`w->window()->setCursor(Qt::BlankCursor);`

tbertels:
Because of a KDE Plasma 6.2.x bug, the mpvwidget doesn't extend to the rightmost pixels of the window with screen scaling, so we have to hide the cursor in the window too.
This has side effects in windowed mode, so we only do it in fullscreen mode.
Fixes #294.

### DIFF
--- a/mpvwidget.cpp
+++ b/mpvwidget.cpp
@@ -641,8 +641,10 @@ void MpvObject::setMpvOptionVariant(QString name, QVariant value)
 
 void MpvObject::showCursor()
 {
-    if (widget)
+    if (widget) {
         widget->self()->setCursor(Qt::ArrowCursor);
+        widget->self()->window()->setCursor(Qt::ArrowCursor);
+    }
 }
 
 void MpvObject::hideCursor()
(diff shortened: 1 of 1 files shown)


tbertels:
With this backported, we should be ready for release.

Issue #128, opened 2024-05-24
Would you consider building AppImage in Ubuntu 20.04?


stdedos:
One tiny issue with the AppImages is that their C library is "bound" to the system that they are built against:

```console
...
/tmp/.mount_mpc-qt8DtwPY/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/.mount_mpc-qt8DtwPY/usr/bin/../lib/libpulsecommon-15.99.so)
/tmp/.mount_mpc-qt8DtwPY/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/.mount_mpc-qt8DtwPY/usr/bin/../lib/libpulsecommon-15.99.so)
/tmp/.mount_mpc-qt8DtwPY/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/.mount_mpc-qt8DtwPY/usr/bin/../lib/libdecor-0.so.0)
/tmp/.mount_mpc-qt8DtwPY/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/.mount_mpc-qt8DtwPY/usr/bin/../lib/libkrb5.so.3)
...
$ ldd --version ldd
ldd (Ubuntu GLIBC 2.31-0ubuntu9.15) 2.31
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

I haven't found you calling https://github.com/mpc-qt/mpc-qt/blob/master/make-release-linux.sh on the CI, so, I assume you build manually/locally? (I guess that makes it harder to build against multiple systems :sweat_smile: )

cmdrkotori:
>Would you consider building AppImage in Ubuntu 20.04?

No. There are no Qt6 packages on 20.04.  There is also the matter that [Qt6 is not supported on 20.04](https://doc.qt.io/qt-6/linux.html).

fwiw, the last Qt 5 version was commit df7942655b1ba645a4cc2b14a1a739d655cc44d8.  You can try building that on your system.

[...]

PR #297, opened 2024-12-07
Build script updates

cmdrkotori:
These are the last minute changes to the build scripts I had to make today when building the latest release.  As you can see, pretty benign stuff.

### DIFF
--- a/make-release-linux.sh
+++ b/make-release-linux.sh
@@ -59,7 +59,7 @@ wget https://github.com/linuxdeploy/linuxdeploy-plugin-qt/releases/download/cont
 chmod +x linuxdeploy*.AppImage
 
 # initialize AppDir, bundle shared libraries, add desktop file and icon, use Qt plugin to bundle additional resources, and build AppImage, all in one command
-EXTRA_QT_PLUGINS="svg;" QMAKE=/usr/bin/qmake6 ./linuxdeploy-x86_64.AppImage --appdir AppDir -e mpc-qt -i "$REPO_ROOT"/images/icon/mpc-qt.svg -d "$REPO_ROOT"/mpc-qt.desktop --plugin qt --output appimage
+EXTRA_QT_PLUGINS="svg;" QMAKE=/usr/bin/qmake6 ./linuxdeploy-x86_64.AppImage --appdir AppDir -e bin/mpc-qt -i "$REPO_ROOT"/images/icon/mpc-qt.svg -d "$REPO_ROOT"/mpc-qt.desktop --plugin qt --output appimage
 
 # move built AppImage back into original CWD
 mv Media_Player_Classic_Qute_Theater-x86_64.AppImage "$OLD_CWD/$DEST.AppImage"
--- a/make-release-win.sh
+++ b/make-release-win.sh
@@ -62,7 +62,7 @@ platforms/qdirect2d.dll
 platforms/qminimal.dll
 platforms/qoffscreen.dll
 platforms/qwindows.dll
-styles/qwindowsvistastyle.dll
+styles/qmodernwindowsstyle.dll
 EOF
 
 read -r -d '' docs <<'EOF'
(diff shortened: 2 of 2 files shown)


cmdrkotori:
This lacks the project file edit of replacing the line `unix:PKGCONFIG += mpv` with `unix:LIBS += -lmpv` for the [unique build environment](https://github.com/mpc-qt/mpc-qt/issues/128#issuecomment-2131177394) required to build the AppImage on Ubuntu 20.04.

tbertels:
On the subject, is there any reason to have both `make-release-msys2.sh` and `make-release-win.sh` or could the release be made by the GitHub action?

For the AppImage, we could probably automate it too if we build or download newer dependencies than on the Ubuntu system.
I just saw your comment, we can force use ubuntu-24.04 before it becomes the default this month.

cmdrkotori:
>On the subject, is there any reason to have both make-release-msys2.sh and make-release-win.sh or could the release be made by the GitHub action?

The big reason why make-release-win is used is because it results in a smaller directory listing due a static build of libmpv, and it bundles yt-dlp.  The second reason is I have no experience in making releases from Github actions.

>For the AppImage, we could probably automate it too if we build or download newer dependencies than on the Ubuntu system.
>I just saw your comment, we can force use ubuntu-24.04 before it becomes the default this month.
>Update: It seems like we actually want the opposite: force use ubuntu-20.04.

I see no reason not to do this.

tbertels:
> > On the subject, is there any reason to have both make-release-msys2.sh and make-release-win.sh or could the release be made by the GitHub action?
> 
> The big reason why make-release-win is used is because it results in a smaller directory listing due a static build of libmpv, and it bundles yt-dlp. The second reason is I have no experience in making releases from Github actions.

I'm sure we could download automatically the libmpv and yt-dlp builds like you do it manually for the release. We could even host a backup of them in a different repository if needed.
I can work on it to make the future releases easier.

> > For the AppImage, we could probably automate it too if we build or download newer dependencies than on the Ubuntu system.
> > I just saw your comment, we can force use ubuntu-24.04 before it becomes the default this month.
> > Update: It seems like we actually want the opposite: force use ubuntu-20.04.
> 
> I see no reason not to do this.

Do you mean you don't see a reason to automate the AppImage or the part about the Ubuntu versions?

I see the AppImage as a way for users to be able to use mpc-qt even if they don't a way to automatically build it.
The executable artifact requires the right dependencies, so it's not actually that useful for users.

I was thinking of adding links to the latest continuous builds on the website like mpv does, but by using https://github.com/TheAssassin/pyuploadtool to upload the latest artifacts in a new repository and then link them from the website.
Because the users need to have an account to download the CI artifacts and it's not very user friendly anyway.

cmdrkotori:
>I'm sure we could download automatically the libmpv and yt-dlp builds like you do it manually for the release. We could even host a backup of them in a different repository if needed.
>I can work on it to make the future releases easier.

Thanks. I'm not always at home with my rig to do manual builds as it is.

>Do you mean you don't see a reason to automate the AppImage or the part about the Ubuntu versions?

Sorry, I meant automating the appimage.  Actions can be on whatever ubuntu version the situation calls for.

>I see the AppImage as a way for users to be able to use mpc-qt even if they don't a way to automatically build it.
>The executable artifact requires the right dependencies, so it's not actually that useful for users.

Yes I agree.  Luckily we already have a make-release-linux script that can be cannibalized to do the appimage part. (finding the documentation for it wasn't easy.)

>I was thinking of adding links to the latest continuous builds on the website like mpv does, but by using https://github.com/TheAssassin/pyuploadtool to upload the latest artifacts in a new repository and then link them from the website.
>Because the users need to have an account to download the CI artifacts and it's not very user friendly anyway.

Sounds good.  The more testers the better.

>About the website, we probably want to replace the main screenshot with the one of the readme, even if we don't update the gallery yet.

I'll get to that once the release is out.

cmdrkotori:
Oh, do you want to me to continue making the next release manually or wait until it's done by CI?

tbertels:
> Yes I agree. Luckily we already have a make-release-linux script that can be cannibalized to do the appimage part. (finding the documentation for it wasn't easy.)

Yes, and your build instructions for Ubuntu 20.04 will be most useful.

> Oh, do you want to me to continue making the next release manually or wait until it's done by CI?

It's probably better if you do it manually this time as I don't know how long it'll take to prepare the CI or if it'll work at all.

tbertels:
> > About the website, we probably want to replace the main screenshot with the one of the readme, even if we don't update the gallery yet.
> 
> I'll get to that once the release is out.

Nice, you've updated the favicon too! 🙂

PR #304, opened 2024-12-10
Add a setting to remember panels state and keep last window geometry setting active

### LINKED ISSUE #301: Save Window Layout
I used to use MPC-HC with minimal layout.
Every time I open MPC-QT, it resets to normal layout.

tbertels:
- Add a setting to remember panels state
This allows saving the panels state without saving the window geometry setting.
This also always remembers the window maximized state.

In Wayland mode, 4c058600b4e9760e2908081997962282fa5693ed disabled the geometry settings.
This means that these settings can't be changed, but they were still being followed for the panels and maximized state.

Remembering the playlist state still requires the auto zoom to be enabled though, otherwise it's always hidden on start.

- Don't disable "Remember last window geometry" in Wayland mode
This isn't the right culprit for the "part of the window can get hidden when opening video in Wayland mode" bug. Or it isn't anymore.

Fixes #301.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -873,6 +873,8 @@ void Flow::setupFlowConnections()
             this, &Flow::settingswindow_rememberFilePosition);
     connect(settingsWindow, &SettingsWindow::rememberWindowGeometry,
             this, &Flow::settingswindow_rememberWindowGeometry);
+    connect(settingsWindow, &SettingsWindow::rememberPanels,
+            this, &Flow::settingswindow_rememberPanels);
     connect(settingsWindow, &SettingsWindow::mprisIpc,
             this, &Flow::settingswindow_mprisIpc);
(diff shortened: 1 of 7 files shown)


tbertels:
We shouldn't disable the "Remember last window geometry" in Wayland mode.
First of all, even when disabled, the window geometry will be saved if the checkbox is checked.
And since Wayland never hides part of the window on start, it's not a problem.

What's a problem is the auto zoom setting, especially in its default 100% setting.
This means that if mpc-qt starts with a small centered window and the user plays a file near the screen resolution, the window will get resized to 100% and half the window will be hidden.

Disabling auto zoom fixes that problem.
But there's a bug where the video isn't properly fitted to inside of the window until either the playlist is shown /hidden, the window is resized or switched to fullscreen. And this happens too if the window is maximized on start.

PR #313, opened 2024-12-16
Update subtitles icons

cmdrkotori:


tbertels:
Nice!

The icons don't seem to be centered in their frame.
This was already the case with the old ones, but I'm thinking it wouldn't hurt to center them while you're at it.

tbertels:
And shouldn't they be slightly bigger?
The old ones represented the subtitles but the new ones are supposed to show a screen, and they look smaller than the speaker icon next to it.

cmdrkotori:
>The icons don't seem to be centered in their frame.

Yeah I just checked kde's guidelines for icons and they have everything centered.

Why did this happen? Back in the day when I was writing Win32 apps, small icons in toolbars typically had their bottom-most and rightmost pixels blank, so everything was up and to the left slightly.

>And shouldn't they be slightly bigger?

Possibly yes. If the margins don't have to be as big, the visible part of the icon can be bigger.  I checked the icons in /usr/share/icons/breeze/actions/16, and they seem to have 1px (sometimes) or 2px (most) margins all around the sides.  It seems I've used 2+3 pixel margins so far.

>By the way, we probably want to revisit all the icons to center them some time in the future.

This will probably take anywhere up to a few hours as all the icons have to be edited.  I might get to it sometime in the coming week.

tbertels:
> Why did this happen? Back in the day when I was writing Win32 apps, small icons in toolbars typically had their bottom-most and rightmost pixels blank, so everything was up and to the left slightly.

Wasn't it because the buttons click animation used to be them being pushed down? And that animation moved the icon down and to the right, so there had to be some free space.

PR #320, opened 2024-12-20
Make icons images larger

cmdrkotori:
Make fallback icon theme images larger.  Images are centered and take the full width and height of the image area.

Follow up to #313

Screenshot_20241220_192556


tbertels:
Does this preserve the aspect ratio of all buttons icons?
The pause button seems to be thinner compared to its height.

PR #330, opened 2025-01-06
Support pan and scan

### LINKED ISSUE #55: Pan and scan

tbertels:
This makes it possible to for example correctly display on a 16:9 screen a 16:9 video made for 4:3 screens with letterbox black bars encoded in the video.
Since there are enough entries, it's also time to move them to submenus.
Fixes #55.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1989,10 +1989,16 @@ void MainWindow::setVideoTracks(QList<Track> tracks)
         return;
     ui->menuPlayVideo->setEnabled(true);
     videoTracksGroup = new QActionGroup(this);
-    ui->menuPlayVideo->addAction(ui->actionDecreaseVideoAspect);
-    ui->menuPlayVideo->addAction(ui->actionIncreaseVideoAspect);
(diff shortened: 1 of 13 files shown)


cmdrkotori:
Sorry for how long this took to get to.  The family is dealing with a medical emergency and I'm not sure when I'll have the opportunity to check github on a timely basis.  Perhaps for the moment you can merge pull requests yourself.

tbertels:
I'm really sorry to hear that, I hope it'll turn out alright.

No problem, I'm actually rather busy too right now.

PR #351, opened 2025-02-13
Add translations

tbertels:
Tamil is 100% complete and Catalan is at 90%.

There seems to be a problem with Bokmål (a Norwegian language): Weblate says that it's 100% complete with zero string.
Is there a way to remove it or reset it?
Note that it only shows up on https://hosted.weblate.org/projects/mpc-qt/#languages


cmdrkotori:
>Is there a way to remove it or reset it?

Not that I know of.  Weblate is usually a hands-off process that solves itself; I haven't had to do much admin after setting it up.  Curiously enough, the translation doesn't show up at https://hosted.weblate.org/projects/mpc-qt/mpc-qt-translations/, so it would seem to be that the translator started a translation on the glossary page rather than the translations page.

PR #356, opened 2025-02-21
Add an AppImage action

### LINKED ISSUE #308: Appimage hardware decoding doesn't work
Hi, version 24.12. This isn't libmpv's fault because hardware decoding works in other players (as well as Haruna which also uses libmpv). You can see that the CPU usage is higher despite the checkbox being checked.

All the other players I am using are flatpaks.. mpc-qt is the only appimage one, so I am not sure if this is playing a role or not.

I would also suggest introducing a key to allow people to toggle hw decoding on and off like how mpv uses ctrl-h to do so, and you can see the output of that reflected in the OSD.

Operating System: SteamOS 3.6.20 Build: 20241030.1
KDE Plasma Version: 5.27.10
KDE Frameworks Version: 5.115.0
Qt Version: 5.15.12
Kernel Version: 6.5.0-valve22-1-neptune-65-g9a338ed8a75e (64-bit)
Graphics Platform: X11

tbertels:
Uses build instructions for Ubuntu 20.04 from #128

It may fix #308 (no hardware decoding with AppImage) with the custom AppRun though it uses libva-copy instead of libva here.

Uses a VERSION file to make it easier to set the correct bugfix .X version on a release branch.


tbertels:
The AppImage file is much bigger most likely because of the included libLLVM and libgallium libraries.

tbertels:
I forgot to mention: for some reason, the recent versions of linuxdeploy end up in an unusable AppImage.
They don't seem to use a valid SquashFS format.

Issue #31, opened 2022-03-01
Publish as a flatpak


SuNNjek:
Hello,
seeing as mpv has been [published as a flatpak](https://flathub.org/apps/details/io.mpv.Mpv), would it be possible to get mpc-qt published as one as well? This would make installing mpc-qt without compiling it yourself possible on any distro that supports flatpaks (which is most if not all at this point). Having a bit of experience with flatpaks, I would be happy to assist.

ahjolinna:
also isn't OSTree (Fedora's Silverblue/Kinoite, SteamOS v3 and SUSEs MicroOS)  becoming more relevant which means flatpak would be needed more for these 3rd-party apps

ahjolinna:
@SuNNjek I was wondering what is flathub/flatpaks policy on some codecs or/and supporting stuff like CUDA, are there some restrictions? (because patents or proprietary)

(I assume mpc-qt need to have its own mpv & maybe ffmpeg)

g-fb:
Here's flatpak manifest (edited from the one for my own player). The app runs, but playback fails.
You build and run it with `flatpak-builder build com.github.mpc-qt.yml mpc-qt --force-clean --ccache && flatpak-builder --run build com.github.mpc-qt.yml mpc-qt`

Log:
```
[0.555290817] [glwidget] assigning x11 display
[0.636945489] [mpvobject] paused-for-cache property changed, but was not in dispatch list.
[2.647292110] [file] error: Cannot open file '/run/user/1000/doc/817faea5/file.mkv': No such file or directory
[2.647300530] [stream] error: Failed to open /run/user/1000/doc/817faea5/file.mkv.
[2.668615370] [qt] warn: QFSFileEngine::open: No file name specified
```
com.github.mpc-qt.yml:
```yml
---
app-id: com.github.mpc-qt
runtime: org.kde.Platform
runtime-version: '5.15-21.08'
sdk: org.kde.Sdk
command: mpc-qt
rename-icon: mpc-qt
rename-desktop-file: mpc-qt.desktop
copy-icon: true
finish-args:
  - '--share=ipc'
  - '--share=network'
  - '--socket=fallback-x11'
  - '--socket=wayland'
  - '--socket=pulseaudio'
  - '--device=dri'
  - '--filesystem=host'
  - '--talk-name=org.freedesktop.ScreenSaver'
  - '--env=LC_NUMERIC=C'
modules:
  - name: mpc-qt
    buildsystem: qmake
[...]

PR #371, opened 2025-03-10
Add a Flatpak action

tbertels:
Uses the Flatpak YAML from Haruna adapted to KDE 6.7.

#31.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,38 @@
+name: Flatpak maker
+
+on:
+  [workflow_call]
+
+jobs:
+  flatpak:
+    name: Build
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:kde-6.7
+      options: --privileged
+    steps:
+      - name: Check out repository
+        uses: sithlord48/fancy-checkout@v1.0.0
+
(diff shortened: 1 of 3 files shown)


ahjolinna:
shouldn't x264/x265 (and maybe vulkan & vapoursynth) be enabled? at least are in mpv flatpak they are: https://github.com/flathub/io.mpv.Mpv

also I think this is still needed for pipewire to work: `--filesystem=xdg-run/pipewire-0:ro`

tbertels:
> shouldn't x264/x265 (and maybe vulkan & vapoursynth) be enabled? at least are in mpv flatpak they are: https://github.com/flathub/io.mpv.Mpv
> 
> also I think this is still needed for pipewire to work: `--filesystem=xdg-run/pipewire-0:ro`

Good catch, thanks!

As for the FFmpeg `--enable-libx264` and `--enable-libx265` options, [they're actually for encoding](https://trac.ffmpeg.org/wiki/CompilationGuide/Ubuntu#libx264), not decoding.

cmdrkotori:
+1.

The `--enable-libx264` option will be needed if we ever implement the dubious feature to export video clips with mp4 export.  There may be others too, but we can revisit this later.

tbertels:
I've added `--enable-libx264` as disabled for now.

PR #378, opened 2025-03-17
Re-add the appstream file and update it

tbertels:
The appstream file is used by package managers to show information about and screenshots of an application.
Flathub also requires one.
#31

### DIFF
--- a/io.github.mpc_qt.Mpc-Qt.appdata.xml
+++ b/io.github.mpc_qt.Mpc-Qt.appdata.xml
@@ -1,15 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
- <id>io.github.mpc_qt.Mpc-Qt</id>
+ <id>io.github.mpc_qt.mpc-qt</id>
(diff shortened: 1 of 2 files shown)


cmdrkotori:
I notice that the appdata says its id is `io.github.mpc_qt.mpc-qt` while the file is called `io.github.mpc_qt.Mpc-Qt.appdata.xml`, and the desktop file in the code is `io.github.mpc-qt.mpc-qt`.

tbertels:
> I notice that the appdata says its id is `io.github.mpc_qt.mpc-qt` while the file is called `io.github.mpc_qt.Mpc-Qt.appdata.xml`, and the desktop file in the code is `io.github.mpc-qt.mpc-qt`.

I forgot to rename the appstream file. 😕
The id can only contain an hyphen in the last part, the other one has to be converted to an underscore.
About the desktop file, we'll have to rename it again (not just in this file), this time to `io.github.mpc_qt.mpc-qt` like the id. Because since the Flatpak renames it to that too, in Wayland mode the icon bug is back.

cmdrkotori:
No worries, just thought I'd mention it.

Issue #337, opened 2025-01-18
Using KDE Plasma icon does not appear on taskbar when the program is running


flandriai:
Using the desktop environment KDE Plazma with Wayland mpc-qt icon's does not appear on the taskbar. Instead of it the standard Wayland icon appears (Orange circle, with big, white W in it.)

tbertels:
Based on one of your reports, you use mpc-qt on OpenSUSE Tumbleweed.
Some I'm guessing you're using this package: https://build.opensuse.org/package/show/openSUSE%3AFactory/mpc-qt
However, I've noticed that it uses a custom `.desktop` file named `Media Player Classic Qute Theater.desktop`
This may be why there's a problem with the icon, as [the name shouldn't contain spaces](https://specifications.freedesktop.org/desktop-entry-spec/latest/file-naming.html).

How do you start mpc-qt?
Is it pinned to the taskbar?
What kind of taskbar do you use? Icons only or icons and text?

Sunspark-007:
Fedora uses this naming as well.

flandriai:
> Based on one of your reports, you use mpc-qt on OpenSUSE Tumbleweed. Some I'm guessing you're using this package: https://build.opensuse.org/package/show/openSUSE%3AFactory/mpc-qt However, I've noticed that it uses a custom `.desktop` file named `Media Player Classic Qute Theater.desktop` This may be why there's a problem with the icon, as [the name shouldn't contain spaces](https://specifications.freedesktop.org/desktop-entry-spec/latest/file-naming.html).
> 
> How do you start mpc-qt? Is it pinned to the taskbar? What kind of taskbar do you use? Icons only or icons and text?

Yes, I use OpenSUSE Tumbleweed and the mpc-qt built by OpenSUSE team. I always start MPC-QT from Application menu (from the Multimedia section). It is not pinned on the taskbar. The taskbar is the default option provided by OpenSUSE Thumbleweed, but with icons only option. When you start the program for a short while the correct icon does appear (while the program is still being loaded) for a short while on the taskbar, but when the program is loaded the Wayland icon can be seen.

tbertels:
> Fedora uses this naming as well.

As far as I can see, Fedora keeps the `mpc-qt.desktop` name: https://packages.fedoraproject.org/pkgs/mpc-qt/mpc-qt/fedora-42.html

I've tested it in Slowroll KDE with Wayland, the problem happens whenever the desktop file hasn't the same name as the executable.
[...]

PR #380, opened 2025-03-18
Rename .desktop and appstream files

tbertels:
* Rename .desktop file to use recommended appstream format
The id can only contain an hyphen in the last part, the other one has to be converted to an underscore. And since Flatpak renames the .desktop file to the id, we have to follow that format if we want to avoid #337.
Follow-up to fb1a71af709f82f308797d9d9e3a3861a6bd79d2.
* Rename appstream file to use same format as id and .desktop file


cmdrkotori:
The flatpak action failed with the following message

```
Error: Can't load desktop file /__w/mpc-qt/mpc-qt/flatpak_app/files/share/applications/io.github.mpc_qt.mpc-qt.desktop: Error opening file /__w/mpc-qt/mpc-qt/flatpak_app/files/share/applications/io.github.mpc_qt.mpc-qt.desktop: No such file or directory
Error: Build failed: Error: The process '/usr/bin/xvfb-run' failed with exit code 1
```

tbertels:
Yes, that's because it uses (through the manifest file) the version of this repository (where the .desktop file isn't yet renamed) instead of the branch being merged.
The only way to avoid this would have been to keep the `rename-desktop-file` line. But then the action would have failed after the merge.
Hopefully, this will only happen for this specific change.

PR #383, opened 2025-03-19
Update mpc-qt_ja.ts 2025-03-20

coolvitto:


coolvitto:
I have already updated Weblate.
But I requested it because I wanted to sync other files in my fork with the main one, is this not necessary?

tbertels:
No, it's not needed.
Your PR would actually mark the translations strings as `unfinished`, which they're not.
Since you use Weblate, your fork is only useful if you want to test the translations without waiting a few hours for Weblate to update them here.
And Weblate did it one hour ago, so you can test the [latest Windows build](https://github.com/mpc-qt/dev-builds/releases/download/continuous/mpc-qt-continuous-win-x86_64.zip) (this link will always point to the latest build).

tbertels:
No problem! 😉

PR #387, opened 2025-04-01
Switch to the CMake build system

tbertels:
This fixes building the Flatpak version with the KDE 6.8 runtime and ffmpeg 7.0.2 (or ffmpeg-full).
Since Qt recommends CMake, qmake is likely to become less reliable in the long run.
Using CMake, with or without Ninja, provides build progress and speeds up the build.
It also makes it easier to build in a subdirectory to keep the source folder clean.

An .rc file is required to embed the icon and metadata in the Windows build.
Qt doesn't (yet?) provide a way to generate this file with CMake, so I've reused the one created by qmake.

Unlike add_executable, qt_add_executable cannot be called without specifying sources upfront, so we cannot use target_* commands before it.

qt_add_translations adds an update_translations target, so with the command
>cmake -G Ninja -B build ; cmake --build build -t update_translations ; cmake --build build

we can update translations before the build.

* Update Flatpak manifest
Using ffmpeg-full speeds up the build and reduces the Flatpak file size.


tbertels:
Since the master branch isn't yet using CMake, the Flatpak build will fail before this is merged.

cmdrkotori:
Good job, works on my machine.

tbertels:
I'll update the [release/24.12_flatpak](https://github.com/mpc-qt/mpc-qt/tree/release/24.12_flatpak) branch and the Flathub PR.

PR #396, opened 2025-04-15
Remove audio volume from Options and move balance to audio page

### LINKED ISSUE #291: Volume setting isn't synchronized with main window volume slider
When we open the Playback settings, the volume has the same value as during the app start.
This means that when we apply or close the Options with Ok, the main window volume slider gets set to the settings volume.
This is a regression caused by 0e84a7fef2a2eb7bc40618e63a2b3816535503bd.

We could fix this by reverting this commit and just breaking the loop.
But do users really want or need a volume setting in the Options? This seems more like a mirror of mpc-hc, where it didn't make much sense either.

So I'm thinking we could just revert this commit for release, and remove this setting on master.
The balance setting could be moved to the Audio page. The hard part is knowing where to put it exactly.
We could also move balance to the right of After Playback at least for now.

tbertels:
Volume setting isn't very useful in the Options as it is usually set in the main window.
And since 0e84a7fef2a2eb7bc40618e63a2b3816535503bd, when we open the Playback settings, the volume has the same value as during the app start.
This means that when we apply or close the Options with Ok, the main window volume slider gets set to the settings volume.

Fixes #291.


cmdrkotori:
Nice, although the volume is at 100% on every run. The following changes will fix it

In MainWindow::state
```
        { "volume", volumeSlider_->value() }
```

In MainWindow::setState
```
    volumeSlider_->setValue(map.value("volume", 100).toInt());
    volume_sliderMoved(volumeSlider_->value());
```

tbertels:
> Nice, although the volume is at 100% on every run. The following changes will fix it

Good catch!

And of course, I had temporarily removed the `cmake --build build -t update_translations` build command part just before making this PR to compare build speed, so the translations weren't updated. 🤦‍♂️

PR #398, opened 2025-04-18
Add a music mode

### LINKED ISSUE #220: Add an option to disable cover art
- There is an issue where the same cover art is displayed for every track, even when different files have different cover art.
- Due to this bug, MPC-QT cannot be resized to a smaller window, as the large cover art makes it difficult to reduce the size.
- It would be helpful to add an option to disable cover art entirely, so users can use MPC-QT as a music player without cover art display, allowing for a more compact interface.

[cmdrkotori](https://github.com/cmdrkotori) mentioned:
> The bug mentioned in additional information can be fixed by adding gui for the mpv option audio-display somewhere in the settings.

This is a split from #158 by [amayra](https://github.com/amayra).

tbertels:
The purpose is to use all the available space for the playlist and make it possible to keep a small window size.

This enables the playlist, makes it use the full width and height, and prevents resizing, moving or undocking it.
As the playlist uses all the space (apart from the menus and buttons), resizing the window resizes the playlist.
This also disables cover art to prevent the window from getting resized if auto zoom is enabled.
A few buttons are also hidden: step backward and forward, and subtitles.

QApplication::processEvents() is needed to prevent the playlist from playing catch-up with the window on resize.

Fixes #220.

### DIFF
--- a/mainwindow.ui
+++ b/mainwindow.ui
@@ -865,6 +865,7 @@
     <addaction name="actionViewHideLog"/>
     <addaction name="actionViewHideLibrary"/>
     <addaction name="actionViewHideControlsInFullscreen"/>
+    <addaction name="actionViewMusicMode"/>
     <addaction name="menuOSD"/>
     <addaction name="menuViewPresets"/>
     <addaction name="separator"/>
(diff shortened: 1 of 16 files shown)


mainwindow.ui
```diff
@@ -2245,6 +2246,17 @@
     <string>C</string>
    </property>
   </action>
+  <action name="actionMusicMode">
```
cmdrkotori, line 2249: 1 nitpick: it should be `actionViewMusicMode` + update other references.

cmdrkotori (commented):
LGTM, minor nitpick is minor.

tbertels:
I've noticed that the playlist buttons were getting hidden by overflow, so I've added `otherWidgetsHeight` to take the heights of the `menubar` and `bottomArea` into account.

PR #413, opened 2025-05-03
Always use pointing hand cursor for seekbar

tbertels:
If the mouse pointer was on the seekbar before loading the first file, it wouldn't change to a pointing hand cursor even after moving it on the seekbar.
This has to be in MainWindow instead of in DrawnSlider to work.
Follow-up to ac98edd434c56f1f6c35cb07788046ecf4f0fc30.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -3115,6 +3115,8 @@ void MainWindow::position_sliderMoved(int position)
 
 void MainWindow::position_hoverValue(double position, QString chapterInfo, double x)
 {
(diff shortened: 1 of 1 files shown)


cmdrkotori:
Hovering the mouse cursor over the timeline also changes the cursor for the rest of the buttons at the bottom.

tbertels:
> Hovering the mouse cursor over the timeline also changes the cursor for the rest of the buttons at the bottom.

Yes, we do need to restore the arrow cursor on hover end.

PR #415, opened 2025-05-03
Use worst quality for ytdl video preview

### LINKED ISSUE #414: Online video preview support
Please support online video preview, I found online video preview in MPC-BE (this is great!).

tbertels:
This speeds up downloading online videos to show a preview.

Fixes #414.


zhouxinghong:
The best practice for preview quality is to allow users to customize their choices, such as 4K, 2K, 1080P, 720P, etc. If the user's network speed is very fast, then previewing the highest quality will not be too slow. For example, MPC-BE is a very good example.

PR #424, opened 2025-05-07
Support removing subtitles additions for deaf or hard-of-hearing (SDH)

tbertels:
This pretty cool mpv feature removes the subtitles text like "[Screaming]" or "MAN:".
This obviously requires text-based subtitles to work.
Setting in "Subtitles > Misc", disabled by default.
mpv also has a feature to more aggressively remove SDH text and even supports regex filtering, but this should already cover most cases.


tbertels:
We could add support for custom mpv commands like Haruna does.
This would allow us to cover edge cases without cluttering the options.

cmdrkotori:
>We could add support for custom mpv options like Haruna does.
>This would allow us to cover edge cases without cluttering the options.

I see no reason why not.  I stopped keeping up with every mpv commit quite some years ago, so this would allow people to use features before we know about them.

PR #429, opened 2025-05-08
Use Platform::isWindows instead of a macro

tbertels:
Which itself uses a macro, but it's not in the main code.


tbertels:
We do have to use a macro because SIGHUP isn't declared on Windows. 🤦‍♂️

PR #433, opened 2025-05-10
Make more playlist-related strings translatable

tbertels:
Includes an update of the French translation.

### DIFF
--- a/playlistwindow.cpp
+++ b/playlistwindow.cpp
@@ -495,7 +495,7 @@ void PlaylistWindow::newTab()
 {
     bool ok;
     QString title = QInputDialog::getText(this, tr("Enter Playlist Name"),
-                                           "Name", QLineEdit::Normal,
+                                           tr("Name"), QLineEdit::Normal,
                                           tr("New Playlist"), &ok);
     if (!ok)
         return;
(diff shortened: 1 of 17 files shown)


tbertels:
The "Quick playlist" (with that exact spelling) tab title still isn't translated despite `tr("Quick playlist")` being used in `PlaylistCollection::doNewPlaylist`.
It seems that the translator isn't yet initialized when we try to get that translation.

playlistwindow.cpp
```diff
@@ -24,7 +24,7 @@ PlaylistWindow::PlaylistWindow(QWidget *parent) :
     ui->setupUi(this);
     setObjectName("playlistWindow");
     setWindowTitle(tr("Playlist"));
-    addNewTab(QUuid(), tr("Quick Playlist"));
+    addNewTab(QUuid(), tr("Quick playlist"));
```
tbertels, line 27: We may want to instead replace the one occurrence of "Quick playlist" by "Quick Playlist" to keep the same letter case as for "New Playlist".

cmdrkotori: Yeah keep it captialized. 👍

tbertels:
I'd forgotten to rebuild to update the strings.

PR #440, opened 2025-05-18
Never set ToolTip window flag for VideoPreview

tbertels:
It seems to work as well as with Qt::Tool and Qt::FramelessWindowHint so there's no reason to make the code more complex than needed.


tbertels:
The window flags don't make sense anymore if we draw the videopreview as part of the mainwindow itself.

PR #446, opened 2025-05-22
cmakelists: don't pass "-master" to resource compiler

cmdrkotori:
This allows plain cmake builds without a version override to work.


tbertels:
You may want to amend the commit message too.

PR #454, opened 2025-05-30
Output stacktrace in log and handle SIGABRT

tbertels:
Uses addr2line, so it will only output the address on Windows.

On Linux, it will output the function names if it's not a debug build.

On Windows, drmemory's symquery can output function names, but Boost doesn't (yet?) support it.

Killing the app will also output the stacktrace in the log, which is useful in case it hangs.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -45,7 +45,7 @@ jobs:
     - name: Install packages
       run: |
         for i in 1 2 3; do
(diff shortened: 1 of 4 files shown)


tbertels:
We could theoretically ship symquery.exe (and its two required DLLs), and set BOOST_STACKTRACE_ADDR2LINE_LOCATION to the symquery.exe binary in the mpc-qt folder.

But with this we'll now get addresses that we can use with `symquery -e mpc-qt.exe -f -a [address]` to get functions names afterwards.

Issue #52, opened 2022-11-11
Save thumbnails not working....


prime6809:
Hi,

Windows 10 X64, MPC-QT 22.02.

I run mpc-qt and either drop a video file on it or open it from the file menu, all goes fine and the video starts playing, I can seek / skip etc. So nothing wrong with playback.

However if I try and use file->save thumbnails, I get the dialog window where you can set the number/quality and path of the thumbnails ok, but when I click "go" all the controls on the dialog window grey out and nothing seems to happen. I left the video playing till the end but this didn't seem to help. 
The only way I can get out of this is to close mpc-qt.

Is this a known bug or am I doing something wrong?

Cheers.

Phill.

cmdrkotori:
Not a known bug. What file are you trying to do this with?

tbertels:
Does this happen with any file?
Can you reproduce it with https://github.com/mpc-qt/mpc-qt/blob/master/.github/actions/data/test.mkv for example?

PR #460, opened 2025-06-01
Fix Thumbnailer on Windows and center MPC-QT's name vertically

tbertels:
* Don't set thumbState to AvailableState in MpvThumbnailer::mpv_playTimeChanged
At least on Windows, mpv_playTimeChanged can be called with time == -1 instead of 0 when opening a new file.
When mpv_playTimeChanged then gets called again with time == 0, if thumbState != StartedState, the thumbnailing never happens.
Fixes https://github.com/mpc-qt/mpc-qt/issues/52.

* Add more logging in MpvThumbnailer::mpv_playTimeChanged

* Disable audio playback for thumbnailer

* Center MPC-QT's name vertically in thumbnails image
Prevents the bottom of the name from getting cut off. The Q still isn't whole on Windows when Helvetica isn't installed, but at least it doesn't look like a O.

### DIFF
--- a/thumbnailerwindow.cpp
+++ b/thumbnailerwindow.cpp
@@ -259,7 +259,7 @@ void MpvThumbnailer::renderImage()
     QFontMetrics selfMetrics(selfFont);
     QString self = "MPC-QT";
     QRect selfRect = captionArea.adjusted(0, -selfMetrics.descent(),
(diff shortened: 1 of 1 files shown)


cmdrkotori:
Thanks for that. The thumbnailing code's state machine could do with some improvement.

Issue #500, opened 2025-07-23
Shader status


IM71:
Hi,

I want to use my own sharpening shader, but once I loaded it, I don't know where to check if the shader is loaded correctly.
Please help me clarify how to check the status of a shader.

Thank you.

<img width="1280" height="715" alt="Image" src="https://github.com/user-attachments/assets/03392946-97f3-4cda-ab5f-d1ffa4280e05" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/c3a8558a-74fa-4693-a249-19c375a08ea3" />

tbertels:
It should be shown like this in the Frame Timings OSD page (capture taken with mpv `--glsl-shader=` parameter):

<img width="871" height="542" alt="Image" src="https://github.com/user-attachments/assets/654dc286-c2dc-4a49-a290-6293a1f1ddcd" />

But that page doesn't seem to work.

@cmdrkotori That page does work when using vo=gpu, so I guess its linked to the libmpv output. Do you remember when it last worked?
Still, even with vo=gpu, the shaders don't show up in the page.

cmdrkotori:
The frame timings OSD never worked, as mpv doesn't seem to want to accept the stats page change command.

However I can confirm with the film grain shader that gpu shaders are indeed working.

IM71:
Hi,

Could you please provide a simple, known-to-work shader for mpc-qt?
I want to test if I can load the shader.

Thank you.

cmdrkotori:
Sure, I used [filmgrain.glsl](https://raw.githubusercontent.com/haasn/gentoo-conf/xor/home/nand/.mpv/shaders/filmgrain.glsl) to test on a sample of anime (or anything else with flat colors).  Note that some shaders (not this one) require the framebuffer to be in floating point.

IM71:
> Sure, I used [filmgrain.glsl](https://raw.githubusercontent.com/haasn/gentoo-conf/xor/home/nand/.mpv/shaders/filmgrain.glsl) to test on a sample of anime (or anything else with flat colors). Note that some shaders (not this one) require the framebuffer to be in floating point.

Thank you, confirmed working, both your shader and my sharpening shader.

So, there's no way to check if the shader is working besides eyeball it, right?

cmdrkotori:
>So, there's no way to check if the shader is working besides eyeball it, right?

Correct.

IM71:
[...]

PR #505, opened 2025-07-28
Frame Timings -> Input Cache Stats

cmdrkotori:
Replace Frame Timings with Input Cache Stats.  We can't use frame timings page as the page is blank, but there is some interesting data on other pages.

As per discussion on #500.


mpvwidget.cpp
```diff
@@ -1263,8 +1263,8 @@ void MpvController::setLogLevel(QString logLevel)
 
 void MpvController::showStatsPage(int page)
 {
-    bool statsVisible = (shownStatsPage > 0 && shownStatsPage < 3);
-    bool wantVisible = (page > 0 && page < 3);
+    bool statsVisible = (shownStatsPage > 0 && shownStatsPage <= 5);
+    bool wantVisible = (page > 0 && page <= 5);
```
tbertels, line 1267: Is there still a point in checking if shownStatsPage and page are less than the max page (6 now)? 2 would keep it "visible" anyway, and it would be as blank as with 6.

mainwindow.ui
```diff
@@ -526,7 +526,7 @@
                 <string>Mute</string>
                </property>
                <property name="icon">
-                <iconset resource="build/.qt/rcc/res.qrc">
+                <iconset resource="build/Desktop6-Debug/.qt/rcc/res.qrc">
```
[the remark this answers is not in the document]
cmdrkotori: It's a leftover of editing the file in QtCreator.  It doesn't materially affect the resulting executable.  In any case I'll amend the commit so that it points to the plain cmake folder.

mpvwidget.cpp
cmdrkotori: Rereading your comment, yes, I suppose we could drop the `<= 5` expression.

tbertels:
`MpvObject::cycleStatsPage()` needs to be fixed.
Right now, the Cycle [OSD] command still goes from `-1` to `2`.

PR #506, opened 2025-07-28
Render tinyicon in status bar correctly with screen scaling

tbertels:
* mainwindow: Render tinyicon in status bar correctly with screen scaling
Qt doesn't take into account the device pixel ratio when rendering it automatically.
* mainwindow: Use a constexpr for the tray icon path


tbertels:
Here's before and after:

Copie d'écran_20250728_215435

Copie d'écran_20250728_215814

PR #508, opened 2025-07-29
Improve stats page cycling

cmdrkotori:
Introduce a stats page lookup table in MpvWidget::cycleStatsPage to correctly skip over page 2.  Introduce another lookup table in MainWindow::on_actionViewOSDCycle_triggered to map returned pages to their corresponding actions.

Follow up to #505.

### DIFF
--- a/translations/mpc-qt_ar.ts
+++ b/translations/mpc-qt_ar.ts
@@ -1013,10 +1013,6 @@
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>&amp;Frame Timings</source>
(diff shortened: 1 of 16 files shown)


tbertels:
You may want to add `cmake --build build -t update_translations` to your build command to update the translations.

cmdrkotori:
Ah shoot, I'm too used to the old way where translations were updated on every build. EDIT: fixed it, was a simple checkbox under project options.

tbertels:
I've just noticed that `osdActionGroup` isn't used anymore and generates a build warning.

PR #519, opened 2025-08-05
Add support for custom mpv options

tbertels:
In Tweaks, options entered as option_name=value with a space between different options.
These custom options override all other options with the same name.
I haven't found a way to left align the checkbox and text (same thing for the "Show time tooltip" and "Change OSD font" settings).


cmdrkotori:
>I haven't found a way to left align the checkbox and text (same thing
for the "Show time tooltip" and "Change OSD font" settings).

We could change the page to a grid layout, but labels wouldn't be on the RHS on right-to-left languages.  I guess that's a small price to pay for getting it right most of the time.

PR #520, opened 2025-08-05
mainwindow: Use 1 to 20 numbering for recent files

tbertels:
This is purely cosmetic, but users probably start counting from 1 instead of 0.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1721,7 +1721,7 @@ void MainWindow::setRecentDocuments(QList<TrackInfo> tracks)
     for (int i = 0; i < tracks.count() && i < 20; i++) {
         TrackInfo track = tracks[i];
         QString displayString = track.url.fileName();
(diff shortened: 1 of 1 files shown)


tbertels:
Is it actually useful to number the entries? It doesn't seem to add any information.

SMPlayer numbers them, but it also uses that number for keyboard navigation, which we don't.

cmdrkotori:
>Is it actually useful to number the entries? It doesn't seem to add any information.

It seems to be a desktop tradition to number the recent entries, so no real reason.

tbertels:
> > Is it actually useful to number the entries? It doesn't seem to add any information.
> 
> It seems to be a desktop tradition to number the recent entries, so no real reason.

VLC and SMPlayer do it, but neither MPC-HC nor Haruna.

cmdrkotori:
In that case perhaps the better thing to do is to drop the numbering entirely so we can be closer to MPC-HC.

PR #523, opened 2025-08-06
Add 'Move File to Trash' feature

### LINKED ISSUE #374: [Feature Request] Add shortcut to delete file
I can delete current play file by pressing 'del' on keyboard in mpc-hc/be, but not anymore in mpc-qt.

zakius:
This commit adds a new feature that allows users to delete the currently playing file from the filesystem and automatically move to the next playable file in the same directory.

Key features:
- New action 'Delete Current File and Next'
- Works with both direct files and playlist items
- Uses existing playNextFileUrl() logic for consistency
- Safe deletion: finds next file first, then deletes current
- HTTP remote control support (command 921)
- Appears in Settings → Keys for custom binding

Files modified:
- mainwindow.ui: Added action definition and menu integration
- mainwindow.h/cpp: UI integration and signal/slot handling
- manager.h/cpp: Core file deletion and next file logic
- main.cpp: Signal connections between components
- ipc/http.h/cpp: HTTP command support

This addresses the user request for a command that removes the currently played file and moves to the next one, even when only the playback window is active.

References #374

### DIFF
--- a/ipc/http.cpp
+++ b/ipc/http.cpp
@@ -458,6 +458,7 @@ void MpcHcServer::setupWmCommands()
         { 894, "Decreate Rate",             [&](){ emit decreaseRate(); } },
         { 920, "Next File",                 [&](){ emit nextFile(false); } },
         { 919, "Previous File",             [&](){ emit previousFile(false); } },
+        { 921, "Move to Recycle Bin",       [&](){ emit moveToRecycleBin(); } },
         { 975, "Quick add favorite",        [&](){ emit quickAddFavorite(); } },
         { 937, "Organize Favorites...",     [&](){ emit organizeFavorites(); } },
         { 817, "Toggle Caption&amp;Menu",   [&](){ emit toggleCaptionMenu(); } },
--- a/mainwindow.ui
+++ b/mainwindow.ui
@@ -775,6 +775,7 @@
     <addaction name="menuFileOpenDisc"/>
     <addaction name="menuFileRecent"/>
     <addaction name="actionFileClose"/>
+    <addaction name="actionFileMoveToRecycleBin"/>
     <addaction name="separator"/>
     <addaction name="actionFileSaveCopy"/>
     <addaction name="actionFileSaveImage"/>
@@ -1113,6 +1114,14 @@
    <property name="shortcut">
     <string>Ctrl+W</string>
    </property>
+  </action>
+  <action name="actionFileMoveToRecycleBin">
+    <property name="text">
+      <string>Move File to Recycle Bin</string>
+    </property>
+    <property name="shortcut">
+      <string notr="true">Del</string>
+    </property>
   </action>
   <action name="actionFileSaveCopy">
    <property name="text">
--- a/manager.cpp
+++ b/manager.cpp
@@ -1,4 +1,6 @@
 #include <QRegularExpression>
+#include <QFile>
+#include <QFileInfo>
 #include "manager.h"
 #include "playlistwindow.h"
 #include "logger.h"
@@ -324,13 +326,12 @@ void PlaybackManager::navigateToPrevChapter()
         playPrev(false);
 }
 
-void PlaybackManager::playNext(bool forceFolderFallback)
+bool PlaybackManager::playNext(bool forceFolderFallback)
 {
-    if ((folderFallback || forceFolderFallback) && playlistWindow_->isPlaylistSingularFile(nowPlayingList)) {
-        playNextFile();
-    } else {
-        playNextTrack();
-    }
+    if ((folderFallback || forceFolderFallback) && playlistWindow_->isPlaylistSingularFile(nowPlayingList))
+        return playNextFile();
+    else
+        return playNextTrack();
 }
 
 void PlaybackManager::playPrev(bool forceFolderFallback)
@@ -340,6 +341,39 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
     } else {
         playPrevTrack();
     }
+}
+
+void PlaybackManager::moveToRecycleBin()
+{
+    if (nowPlayingItem.isNull())
+        return;
+
+    QUuid oldPlayingItem = nowPlayingItem;
+    QUrl currentFileUrl;
+    if (nowPlayingList.isNull())
+        currentFileUrl = nowPlaying_;
+    else
+        currentFileUrl = playlistWindow_->getUrlOf(nowPlayingList, nowPlayingItem);
+
+    if (currentFileUrl.isEmpty() || !currentFileUrl.isLocalFile())
+        return;
+
+    QString filePath = currentFileUrl.toLocalFile();
+    QFileInfo fileInfo(filePath);
+
+    bool hasNextFile = playNext(folderFallback);
+    emit removePlaylistItemRequested(oldPlayingItem);
+
+    if (fileInfo.exists()) {
+        QFile file(filePath);
+        if (file.moveToTrash())
+            mpvObject_->showMessage(tr("File moved to recycle bin: %1").arg(fileInfo.fileName()));
+        else
+            mpvObject_->showMessage(tr("Failed to move file to recycle bin: %1").arg(fileInfo.fileName()));
+    }
+
+    if (!hasNextFile)
+        stopPlayer();
 }
 
 void PlaybackManager::repeatThisFile()
@@ -780,7 +814,7 @@ void PlaybackManager::updateChapters()
     emit chaptersAvailable(list);
 }
 
-void PlaybackManager::playNextTrack()
+bool PlaybackManager::playNextTrack()
 {
     PlaylistItem next;
     next = playlistWindow_->getItemAfter(nowPlayingList, nowPlayingItem);
@@ -790,8 +824,11 @@ void PlaybackManager::playNextTrack()
         next = playlistWindow_->getFirstItem(nowPlayingList);
     }
     QUrl url = playlistWindow_->getUrlOf(next.list, next.item);
-    if (!url.isEmpty())
+    if (!url.isEmpty()) {
         startPlayWithUuid(url, next.list, next.item, false);
+        return true;
+    }
+    return false;
 }
 
 void PlaybackManager::playPrevTrack()
@@ -836,10 +873,10 @@ bool PlaybackManager::playNextFileUrl(QUrl url, int delta)
     return true;
 }
 
-void PlaybackManager::playNextFile(int delta)
+bool PlaybackManager::playNextFile(int delta)
 {
     QUrl url = playlistWindow_->getUrlOfFirst(nowPlayingList);
-    playNextFileUrl(url, delta);
+    return playNextFileUrl(url, delta);
 }
 
 void PlaybackManager::playPrevFile()
--- a/playlistwindow.cpp
+++ b/playlistwindow.cpp
@@ -832,6 +832,16 @@ void PlaylistWindow::playlist_removeItemRequested()
     updatePlaylistHasItems();
 }
 
+void PlaylistWindow::removePlaylistItem(const QUuid &itemUuid)
+{
+    auto qdp = currentPlaylistWidget();
+    if (!qdp)
+        return;
+
+    qdp->removeItem(itemUuid);
+    updatePlaylistHasItems();
+}
+
 void PlaylistWindow::playlist_removeAllRequested()
 {
     auto qdp = currentPlaylistWidget();
--- a/translations/mpc-qt_ar.ts
+++ b/translations/mpc-qt_ar.ts
@@ -1400,6 +1400,10 @@
         <source>&amp;Compressor</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Move File to Recycle Bin</source>
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>MouseState</name>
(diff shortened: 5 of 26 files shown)


manager.cpp
```diff
@@ -342,6 +344,65 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
...
+        }
+    } else {
+        if (playNextFileUrl(currentUrl, 1)) {
+            hasNextFile = true;
+        }
+    }
+
+    if (fileInfo.exists()) {
+        QFile file(filePath);
+        if (file.remove()) {
```
tbertels, line 394: It should use QFile::moveToTrash().

tbertels:
If you can use the command
`cmake -G Ninja -B build ; cmake --build build -t update_translations ; cmake --build build`
to build, the `cmake --build build -t update_translations` will update the translation strings so that you can add them.

translations/mpc-qt_ar.ts
```diff
@@ -1400,6 +1400,14 @@
         <source>&amp;Compressor</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Move to Recycle Bin</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>F12</source>
```
cmdrkotori, line 1408: Don't mark your shortcut as translatable.

ipc/http.cpp
```diff
@@ -458,6 +458,7 @@ void MpcHcServer::setupWmCommands()
         { 894, "Decreate Rate",             [&](){ emit decreaseRate(); } },
         { 920, "Next File",                 [&](){ emit nextFile(false); } },
         { 919, "Previous File",             [&](){ emit previousFile(false); } },
+        { 921, "Move to Recycle Bin", [&](){ emit moveToRecycleBin(); } },
```
cmdrkotori, line 461: Spacing

mainwindow.ui
```diff
@@ -1628,6 +1629,14 @@
...
    </property>
   </action>
+                    <action name="actionNavigateFilesMoveToRecycleBin">
+                   <property name="text">
+                    <string>Move to Recycle Bin</string>
+                   </property>
+                   <property name="shortcut">
+                    <string>F12</string>
+                   </property>
+                  </action>
```
tbertels, line 1639: The command should be moved (and the action should be renamed) from the Navigate menu to the File menu, below "Close File".

translations/mpc-qt_ar.ts
tbertels: In case you're wondering, it just means adding `notr="true"` to the string tag.

mainwindow.ui
tbertels: The command should actually say "Move File to Recycle Bin", since it's not a contextual menu and to mirror the "Close File" command.

mainwindow.ui
```diff
@@ -1628,6 +1629,14 @@
...
    </property>
   </action>
+  <action name="actionFileMoveToRecycleBin">
+    <property name="text">
+      <string>Move File to Recycle Bin</string>
+    </property>
+    <property name="shortcut">
+      <string notr="true">F12</string>
+    </property>
+  </action>
```
tbertels, line 1639: This should be moved in the .ui file after actionFileClose.

tbertels:
After more testing, I think the next file in the current playlist should always be played next if there's more than one file in it, even in the quick playlist.

If we're in a custom playlist (e.g. not the quick playlist) and we move the current file to trash:
- if it's not the last item, the current item becomes blank but stays in the playlist.
- if it's the last item, the item stays in the playlist.

tbertels:
I think the way to fix the playlist item removal is to add a function similar to `PlaylistWindow::playlist_removeItemRequested()` in playlistwindow.cpp to remove the current item (`qdp->currentItemUuid()`).
And then call that function from manager by sending a signal to playlistwindow and connecting that signal in main.cpp.

tbertels:
To fix the part about always playing the next item if there's more than one item in the playlist, you should:
- use `playNext(folderFallback)`. This also makes sure that the corresponding Tweaks option is followed.
- modify `playNext()` (and both `playNextFile()` and `playNextTrack()`) so that they return `true` when there's a next file to be played, `false` otherwise. `playNextFileUrl()`, called by `playNextFile()`, already does that.
- then if `playNext(folderFallback)` is false, call
```
mpvObject_->setPaused(false);
mpvObject_->stopPlayback();
```
About the coding style, don't use curly braces for one line.

zakius:
I think I managed to get all your notes, hopefully there are no more edge cases left

mainwindow.ui
```diff
@@ -1114,6 +1115,14 @@
     <string>Ctrl+W</string>
    </property>
   </action>
+  <action name="actionFileMoveToRecycleBin">
+    <property name="text">
+      <string>Move File to Recycle Bin</string>
+    </property>
+    <property name="shortcut">
+      <string notr="true">F12</string>
```
tbertels, line 1123: The F12 key shortcut is already used by the unimplemented "Export Encode..." command.
You can remove it from that command so that your new command works.

zakius: Actually the latest release of main project had no default key for this action, can't really check further maintained forks right now as it's quite annoying on Linux, but maybe there shouldn't be a default key in such case?

zakius: the clsid2 fork uses `Del` by default

manager.cpp
```diff
@@ -342,6 +343,51 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
...
+{
+    if (!nowPlayingList.isNull() && !nowPlayingItem.isNull())
+        emit removeCurrentPlaylistItemRequested();
+}
+
+void PlaybackManager::requestRemoveCurrentPlaylistItem()
+{
+    removeCurrentPlaylistItem();
+}
+
```
tbertels, line 390: This isn't used anywhere and is redundant with `PlaybackManager::removeCurrentPlaylistItem()`.
You may want to rename `PlaybackManager::removeCurrentPlaylistItem()` to `PlaybackManager::requestRemoveCurrentPlaylistItem()` so that searches in the code don't mix it with `PlaylistWindow::removeCurrentPlaylistItem()`.

tbertels: Actually, you can remove both of these and just use
```
    if (!nowPlayingList.isNull())
        emit removeCurrentPlaylistItemRequested();
```
in `PlaybackManager::moveToRecycleBin()` instead of `removeCurrentPlaylistItem();`

mainwindow.ui
zakius: in that case I'm changing it to `Del`

tbertels:
#374 is actually about both adding a shortcut to remove the current item from playlist and adding a command and shortcut to move the current item to trash.
This currently only deals with the "move the current item to trash" part.
I'm not sure how we want to handle both cases at the same time.
Maybe `Ctrl+Del` to remove from playlist and `Del` to move to trash.

But you don't need to add support for both in this PR.
Though once the PR is ready and you rebase it to one commit, just mention #374 in the commit description (not the title) and without saying "Fixes #374", or it will close that issue by default. We can manually link them afterwards.

tbertels:
You can mention the `Del` key shortcut in the commit description.

manager.cpp
```diff
@@ -342,6 +343,43 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
...
+        QFile file(filePath);
+        if (file.moveToTrash())
+            mpvObject_->showMessage(tr("File moved to recycle bin: %1").arg(fileInfo.fileName()));
+        else
+            mpvObject_->showMessage(tr("Failed to move file to recycle bin: %1").arg(fileInfo.fileName()));
+    }
+
+    if (!hasNextFile) {
+        mpvObject_->setPaused(false);
+        mpvObject_->stopPlayback();
```
tbertels, line 377: These can actually be replaced by `stopPlayer();` and the curly braces won't be needed anymore.
There's still a bug in `stopPlayer()` which doesn't reset the player title, but it's not related to this PR.

manager.cpp
```diff
@@ -342,6 +343,39 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
...
+    if (fileInfo.exists()) {
+        QFile file(filePath);
+        if (file.moveToTrash())
+            mpvObject_->showMessage(tr("File moved to recycle bin: %1").arg(fileInfo.fileName()));
+        else
+            mpvObject_->showMessage(tr("Failed to move file to recycle bin: %1").arg(fileInfo.fileName()));
+    }
+
+    if (!hasNextFile)
+        mpvObject_->stopPlayer();
```
tbertels, line 376: I literally meant `stopPlayer()`, or `PlaybackManager::stopPlayer()` to be clearer, as `MpvObject` doesn't have that function.

tbertels: No problem. 😉
This would have been caught by the build workflow if I didn't have to manually start it each time.

tbertels:
There are still two problems:
- If there are multiple items in the quick playlist, they don't get removed. See review comment
- If we open the second item of a playlist and remove it, then remove the next playing item (the third one), instead of the third one, the first one gets removed. I'll add comments.

manager.cpp
```diff
@@ -342,6 +343,39 @@ void PlaybackManager::playPrev(bool forceFolderFallback)
...
+
+    if (currentFileUrl.isEmpty() || !currentFileUrl.isLocalFile())
+        return;
+
+    QString filePath = currentFileUrl.toLocalFile();
+    QFileInfo fileInfo(filePath);
+
+    bool hasNextFile = playNext(folderFallback);
+    if (!nowPlayingList.isNull())
+        emit removeCurrentPlaylistItemRequested();
```
tbertels, line 365: To fix the second point of https://github.com/mpc-qt/mpc-qt/pull/523#issuecomment-3163412081, you have to :
- add `QUuid oldPlayingItem = nowPlayingItem;` before `QUrl currentFileUrl;` to save the old playing item.
- replace the `removeCurrentPlaylistItemRequested` signal by `removePlaylistItemRequested(oldPlayingItem);`
- (see next comment)

playlistwindow.cpp
```diff
@@ -832,6 +832,18 @@ void PlaylistWindow::playlist_removeItemRequested()
...
+{
+    auto qdp = currentPlaylistWidget();
+    if (!qdp)
+        return;
+
+    QUuid currentItem = qdp->currentItemUuid();
+    if (!currentItem.isNull())
+        qdp->removeItem(currentItem);
+    updatePlaylistHasItems();
+}
```
tbertels, line 845: (continued)
- Rename `PlaylistWindow::removeCurrentPlaylistItem()` to `PlaylistWindow::removePlaylistItem(const QUuid &itemUuid)` - Replace
```
   QUuid currentItem = qdp->currentItemUuid();
   if (!currentItem.isNull())
       qdp->removeItem(currentItem);
```
by
`    qdp->removeItem(itemUuid);`

tbertels: Because after `playNext(folderFallback)`, the current playlist item changes and isn't the one we want to remove anymore.

zakius:
I think I'll may need to stall this a bit, I won't have access to a dev machine for a few days, so if any further issues persist or surface I won't be able to correct them this week

I started it as a "quick fix" to cover my own use case of quickly cleaning up old clips, then tried providing something to the project, but there are many edge cases I haven't considered

tbertels:
> I think I'll may need to stall this a bit, I won't have access to a dev machine for a few days, so if any further issues persist or surface I won't be able to correct them this week

I did test the changes before submitting them, and I haven't found any new ones, so it's all good.

> I started it as a "quick fix" to cover my own use case of quickly cleaning up old clips, then tried providing something to the project, but there are many edge cases I haven't considered

And thank you for that! It's a pretty useful feature to triage media files.

tbertels:
@zakius Thanks again!
I've closed https://github.com/mpc-qt/mpc-qt/issues/374 since it actually was about this exact feature, and I've opened separate issues for the other features mentioned in comments.

tbertels:
I've just noticed that MPC-HC uses the code 921 for `ID_NAVIGATE_SKIPBACK` whereas "Move to Recycle Bin" seems to be 57367 ([`IDS_FILE_RECYCLE`](https://github.com/search?q=repo%3Ampc-hc/mpc-hc%20IDS_FILE_RECYCLE&type=code)).
It also has 24044 ([`ID_FILE_RECYCLE`](https://github.com/search?q=repo%3Ampc-hc%2Fmpc-hc+ID_FILE_RECYCLE&type=code)), but no string for it.

zakius:
I'm sorry for bumping this now but it may be worth noting somewhere that the flatpak releases need rw permissions for filesystem, mine had red only by default
It may be a good idea to change the default permissions too

tbertels:
The flatpak manifest includes these rw permissions:
````
  - --filesystem=/media
  - --filesystem=/run/media
  - --filesystem=/mnt
  - --filesystem=xdg-desktop
  - --filesystem=xdg-documents
  - --filesystem=xdg-download
  - --filesystem=xdg-music
  - --filesystem=xdg-pictures
  - --filesystem=xdg-public-share
  - --filesystem=xdg-videos
````
Where was the file that you tried to remove?

Something that I've noticed with the flatpak is that the files get directly removed instead of being move to the recycle bin.
Qt doesn't use the [XDG Trash Portal](https://github.com/flatpak/xdg-desktop-portal/commit/93f4ec5c6dcea216e2f1da6944aa6894232fe6aa), it's a known Qt bug: https://qt-project.atlassian.net/browse/QTBUG-131744

So I'm thinking of changing the wording to "Remove File" when running the flatpak.
And changing the default shortcut to Shift+Delete for all versions to avoid unintended removals as this feature isn't expected from a media player.

Update: https://github.com/mpc-qt/mpc-qt/wiki/Flatpak#remove-file-doesnt-work

zakius:
it was outside of these but within the home directory

PR #532, opened 2025-08-12
settingswindow: Only enable the Apply button when something can be applied

### LINKED ISSUE #381: The Apply button in settings doesn't show disabled state once it's clicked
As LibreOffice settings window has the same problem, this is an upstream bug.
KDE Plasma 6.2 already had that bug.

tbertels:
We need to ignore keysSearchText when saving settings so that the Apply button doesn't get enabled when filtering keys.

Installing the event filter in the constructor instead of installing /uninstalling it on show / close only triggers it two more times, so it's not worth it to add more code.

Fixes #381.


cmdrkotori:
It was that simple, huh.

PR #536, opened 2025-08-14
Don't cache custom mpv options and resend them when closing Options

tbertels:
Since we set various mpv options when the Options window is closed (in `SettingsWindow::closeEvent`), they potentially erase the custom mpv options.
So we have to set them again when closing Options.
We can't cache them otherwise they only get set the first time we emit options.


settingswindow.cpp
```diff
@@ -1178,6 +1173,16 @@ void SettingsWindow::setFilter(QList<QPair<QString, QString>> &filtersList, QStr
         filtersList.append(QPair<QString, QString>(filter, options));
 }
 
+void SettingsWindow::setCustomMpvOptions()
+{
+    if (ui->tweaksMpvOptionsText->isEnabled()) {
+        QList<MpvOption> mpvOptions = parseMpvOptions(ui->tweaksMpvOptionsText->text());
```
cmdrkotori, line 1179: Am I wrong or should these should use WIDGET_LOOKUP to get the option data from the stored values instead of from the ui?

PR #560, opened 2025-08-28
images: Rename "media-skip-*" and "go-*" buttons icons

### LINKED ISSUE #554: Strange icons at navigation bar
I installed a recent dev version of appImge at openSUSE Leap 15.6.

The interface by default is set like:

Options / Player / Interface / Theme: System (Linux only).

It shows like

Image

Comparing it to [common standard view](https://mpc-qt.github.io/)

The first 3 icons are OK... but then it's getting weird, isn't it?
Skip forward and Skip backward have completly wrong icons
and moreover look like grayed (but are active).
Speed decrease and increase are OK.
Step backward and step forward have wrong icons (double arrow).
Set loop start and end are... well... from out of space :)
and seem grayed (but aren't).

Switching Theme to "Built-in fall back"... phhhh
everything looks like as it shall.

Don't know if it's an issue of MPC-Qt or of the distro etc.

tbertels:
The `sodipodi:docname` are also updated to the new names, but even without that git didn't detect it as files getting renamed.

Follow-up to (and fixes regression introduced by) 2470ff21a80906e63ede3e93b8ddcc1a60784641.

Fixes #554.

### DIFF
--- a/images/theme/black/go-next.svg
+++ b/images/theme/black/go-next.svg
@@ -26,8 +26,8 @@
      inkscape:pageopacity="0.0"
      inkscape:pageshadow="2"
      inkscape:zoom="31.678384"
-     inkscape:cx="7.5130095"
-     inkscape:cy="9.6595836"
+     inkscape:cx="10.638169"
+     inkscape:cy="10.606602"
      inkscape:document-units="px"
      inkscape:current-layer="layer1"
      showgrid="true"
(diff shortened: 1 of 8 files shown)


tbertels:
Some cache prevented me from noticing this regression when I made the PR.
And now that I can see the bug, even after renaming the files, the wrong icons were still shown.
Though with the `sodipodi:docname` update, the icons got updated on my system.

Issue #495, opened 2025-07-21
Create file associations on Windows


tbertels:
Ideally, it would both happen:
- optionally during installation (#494)
There's a [NSIS plugin](https://nsis.sourceforge.io/Application_Association_Registration_plug-in), but the doc isn't that clear and it doesn't mention the file icon. So may be better off adding manually the required registry entries.
- afterwards in the Options (see Player > Formats page).
See also https://wiki.qt.io/Assigning_a_file_type_to_an_Application_on_Windows

An icon for the files would be great, we could embed it in the .exe file as IDI_ICON2.
https://github.com/mpc-qt/mpc-qt/blob/f6f79c36704aab0c7eae126890c06d6610f73b75/platform/mpc-qt.rc.in#L3

https://github.com/he3als/mpc-qt-assoc-install (mentioned in https://github.com/mpc-qt/mpc-qt/issues/27#issuecomment-1874012220) has a nice looking one (with multiple sizes), but I'm not sure how to correctly follow the ISC license used by @he3als.

cmdrkotori:
Another option is to bundle mpc-qt-assoc-install, call it at application install and uninstall, and eschew the page in the settings entirely.

tbertels:
Yes, I didn't want to show a console window, but it looks like we can actually hide it: https://stackoverflow.com/a/45504362/1223073 https://nsis.sourceforge.io/NsExec_plug-in

he3als:
I think you can do anything with the code as long as you add the [license text](https://github.com/he3als/mpc-qt-assoc-install/blob/main/COPYING) somewhere? I would've made it Unlicense personally, but I had to keep the original license as it was forked from https://github.com/rossy/mpv-install

tbertels:
I was going to say that we're supposed to mention the creator of the mpv version, but I see that you've just updated it. 😉
And I better understand now why the readme says 

> Note: This is not necessary if you want to reinstall **mpv** later

Anyway, I've put the license file with the other files in a mpc-qt-assoc sub-folder, so I guess this should do the trick.

PR #587, opened 2025-09-22
settingswindow: Remove "Format" page from Options

tbertels:
7054cdd08e77a941f6aca9cc9efb17f0b7956bf1 added file associations on Windows with the installer and Linux file associations were already set through both the .Desktop file and hints from D-Bus.

With both Linux and Windows having file associations managed by the system, the "Formats" page is unlikely to get implemented in the future.

Suggested in #495.


cmdrkotori:
The parent indicies in on_pageTree_itemSelectionChanged also need to be updated so that clicking items on the left hand side pull up the right page. (decrement every number by 1 except for 0)

PR #589, opened 2025-09-24
settingswindow: Add "Minimum" yt-dlp preset, remove presets, rename setting

### LINKED ISSUE #466: Video freezes or playback suddenly stops after a few seconds or minutes of online video playback of famous bilibili platform
Hi!

So, yt-dl developer explained to me that the famous bilibili platform has special separate audio-only and video-only streams, and I need to install ffmpeg with yt-dlp in order yt-dlp will be able to merge combine unify audio and video :

https://www.bilibili.tv/en/video/4789890817982976?bstar_from=bstar-web.ugc-video-detail.playlist.all

https://www.bilibili.tv/en/video/4789238419162112

This explains why the `worst` preset command format doesn't work in yt-dlp because there are no premerged unified combined streams in bilibili, that's why we should manually put complex ffmpeg commands
to merge audio and video of bilibil of SD 480p video quality of bilibil or test other different presets like :
`-f worstvideo+worstaudio`
because yt-dlp only downloads the best stream out of the box and i prefer SD streams because i have weak wifi and weak old cpu laptop...

All this is very complicated and tiring, so to simplify and save time, I decided to directly playback Bilibili online videos with our beloved great media player which supports yt-dlp= I playback SD video quality from Bilibili. But sadly, often the playback suddenly stops or the video freezes (only audio works) after a few seconds or minutes of online playback of Bilibili videos!

So, we hope this issue will be fixed. Please, thanks!

Because:

China wants to dominate the world and spread its culture, and Bilibili is free for everyone, and some rare anime are only found on Bilibili. Bilibili doesn't have ads, unlike other streaming sites infested with nasty malware ads.
So, it's advantageous to playback anime directly from Bilibili with a media player without a heavy web browser (online video playback with media player = better image quality, better network buffer = smooth playback, we can change aspect ratio and remove black bars, save battery, quiet fan, deinterlace, filters, change codecs etc...).

Thanks!

tbertels:
Users with low bandwith or metered Internet connections need to use the "worstvideo" and "worstaudio" yt-dlp format to get the smallest file possible.
So the new "Minimum" yt-dlp preset does that.
The 240 and 360 presets don't make much sense as users who don't want at least 480 quality probably want the smallest file.
Renaming that setting to "Quality" as that seems more used elsewhere and makes more sense with the "Minimum" preset.

Fixes #466.

### DIFF
--- a/settingswindow.cpp
+++ b/settingswindow.cpp
@@ -161,8 +161,8 @@ QHash<QString, QStringList> SettingMap::indexedValueToText = {
     {"ccHdrMapper", {"clip", "mobius", "reinhard", "hable", "gamma", \
                      "linear"}},
     {"ccHdrCompute", {"auto", "yes", "no"}},
-    {"ytdlpMaxHeight", {"240", "360", "480", "720", "1080", "1440",
-                             "2160", "2880", "4320"}},
+    {"ytdlpQuality", {"minimum", "480", "720", "1080",
+                      "1440", "2160", "2880", "4320"}},
     {"audioChannels", {"auto-safe", "auto", "stereo"}},
     {"audioRenderer", {"pulse", "alsa", "oss", "null"}},
     {"audioAutoloadMatch", { "exact", "fuzzy", "all" }},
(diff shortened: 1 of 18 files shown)


tbertels:
As it is, the "Minimum" preset would always download the smallest height track (for instance 144p).

By using `ytdl-raw-options=format=\"bestvideo[height<=?%1]+bestaudio/best\",format-sort=[res:%1,+size,+br,+res]` instead, we could implement a "Prefer the minimum size" checkbox to download the smallest **filesize** tracks for the given `%1` (for instance 720) video height.
But it doesn't seem worth it to add a setting for that.
Interested users can still use it in the custom mpv options in Tweaks.

PR #609, opened 2025-10-16
Helpers::filterUrls: avoid recursion and circular symlinks

BowDown097:
- Circular symlinks are avoided by checking against a set of visited directories, by their canonical path.
- Recursion is avoided by creating a queue of the URLs, adding items in subdirectories to the queue as needed, with LIFO traversal to preserve the final element order as it was originally. The method used *should* be the most performant way to do so, with some stuff having been changed for it to be possible.
- Also unified and optimized the ``suffix().toLower().split(" ")[0]`` routine.

I think it would also be a good idea to update the filtering routine to use mime types, to be able to hit more files that are perfectly playable, but currently don't pass. I didn't want to make this PR too big, though. Could do that next if there's interest.

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -572,18 +572,14 @@ QList<QUrl> Helpers::filterUrls(QList<QUrl> urls) {
         }
 
         QFileInfo info(u.toLocalFile());
-        QString canonicalPath = canonicalOrAbsolutePath(info);
-
-        if (info.isSymLink()) {
-            QFileInfo linkInfo(info.symLinkTarget());
-            if (canonicalOrAbsolutePath(linkInfo) == canonicalPath) {
-                Logger::logs("helpers", {"skipping circular link:", linkInfo.filePath()});
-                continue;
-            }
+        if (info.isSymLink() && info.absoluteFilePath() == info.canonicalFilePath()) {
+            Logger::logs("helpers", {"skipping circular link:", info.filePath()});
+            continue;
         }
 
+        QString canonicalPath = canonicalOrAbsolutePath(info);
         if (visitedPaths.contains(canonicalPath)) {
-            Logger::logs("helpers", {"skipping likely circular link:", info.filePath()});
+            Logger::logs("helpers", {"skipping visited path:", info.filePath()});
             continue;
         }
 


tbertels:
Could you amend the commit to add the description you added above?

tbertels:
> I think it would also be a good idea to update the filtering routine to use mime types, to be able to hit more files that are perfectly playable, but currently don't pass. I didn't want to make this PR too big, though. Could do that next if there's interest.

Yes, definitely! 
It got mentioned in https://github.com/mpc-qt/mpc-qt/pull/219#issuecomment-2420167763 though we didn't create an issue to track it.

tbertels:
I've tested it with a circular symbolic link, and it doesn't seem to detect it:
- The file still gets added to the playlist.
- I still get the same (translated) error message in the log "Too many levels of symbolic links" and the next file gets played.

I would have expected it to get filtered out like with Haruna.

It would be nice for it to show a message in the log when a circular symbolic link gets filtered out by the way.

BowDown097:
Huh, weird. Maybe I messed up in my testing. I'll try it again.
Edit: 🤦 I only did it for directories. I forgot the same problem would still happen with files. I am absolutely ashamed of me 2 days ago.

BowDown097:
Now this hits every circular symlink case I could think of (entry pointing directly to itself, an entry pointing to a navigation to itself, an entry pointing to an ancestor directory eventually leading to itself).

I also had a question: Why aren't the entries just built up in an unspecified order, then sorted with a collator? This would have a number of benefits:
- Case-insensitive sorting of playlist items
- Natural sorting of playlist items (given the working environment supports it)
- The ability to use QDirIterator over QDir::entryFileList, which is more performant
- (I believe) the ability to abandon the deque-based system introduced here, since it serves mainly as a means to preserve the current ordering

I believe this is what most players do. Is it an intentional choice to not do so here?

tbertels:
It still doesn't work for my simple testcase:

1. I open Dolphin and copy a file in a folder
2. I drag and drop the file to a subfolder and choose to create a symlink
3. I drag and drop the symlink and to the parent folder and choose to create a symlink and overwrite the original file

As far as I can tell, based on what `QFileInfo::absoluteFilePath()` and `QFileInfo::canonicalFilePath()` return, you should:

1. check if `info.isSymbolicLink()`
2. if true, check if `info.absoluteFilePath() == info.canonicalFilePath()`
3. if true, it's a circular symbolic link

tbertels:
Likewise, the same logic can be used for directories.
That way, `Helpers::canonicalOrAbsolutePath` isn't needed.

Then, does deque still make sense or does it now actually slow down access vs how it was before?

Could you also move `Helpers::normalizedSuffix` just after `Helpers::filterUrls`? That way, it's easier to look at the code.

BowDown097:
1. The ``visitedPaths`` logic should still exist for the case that we circle back to the parent directory of whatever files are being opened (or if we're opening a directory, the directory itself). But I think you may be right about ``canonicalOrAbsolutePath`` in retrospect.
2. Technically, yes, but it's essentially a nonexistent difference (I could further explain why). It is afaik by far the best way to produce a filtered version of the data in the same fashion that it is currently done whilst addressing recursion. If the data were to be, say, sorted with a collator after the fact as suggested previously, it would be unnecessary. I have already demoed code that does this locally.
3. :+1:

tbertels:
> 1. The `visitedPaths` logic should still exist for the case that we circle back to the parent directory of whatever files are being opened (or if we're opening a directory, the directory itself). But I think you may be right about `canonicalOrAbsolutePath` in retrospect.

Could you provide simple steps to reproduce, like I did in https://github.com/mpc-qt/mpc-qt/pull/609#issuecomment-3418836018?

>  2. Technically, yes, but it's essentially a nonexistent difference (I could further explain why). It is afaik by far the best way to produce a filtered version of the data in the same fashion that it is currently done whilst addressing recursion. If the data were to be, say, sorted with a collator after the fact as suggested previously, it would be unnecessary. I have already demoed code that does this locally.

I'm asking this because it makes the code more complex. If it doesn't clearly improve performance, simpler code is better.

About your collator suggestion, you're welcome to add a commit to implement it and we'll test it.

BowDown097:
I'd say the most important issue ``visitedPaths`` solves is not navigating in a loop when you have a symlink to an ancestor directory (i.e. directory B in directory A contains a symlink to A). I believe the other issues are pretty minor, enough to perhaps be ignored as a whole for more performance.

And that's good to hear about the collator - it would make the whole deque solution pointless anyway. I can commit the code for it I have locally if you want to have it in this PR. I've tested it on groups of files, directories, and M3U playlists and it seems to function as expected for those.

tbertels:
> I'd say the most important issue `visitedPaths` solves is not navigating in a loop when you have a symlink to an ancestor directory (i.e. directory B in directory A contains a symlink to A). I believe the other issues are pretty minor, enough to perhaps be ignored as a whole for more performance.

I did test that case with two levels of recursion and it worked fine with the logic I suggested in https://github.com/mpc-qt/mpc-qt/pull/609#issuecomment-3418836018:

```
        if (info.isSymLink() && info.absolutePath() == info.canonicalPath()) {
            Logger::logs("helpers", {"skipping circular directory symbolic link:", info.path()});
            continue;
        }
```

> And that's good to hear about the collator - it would make the whole deque solution pointless anyway. I can commit the code for it I have locally if you want to have it in this PR. I've tested it on groups of files, directories, and M3U playlists and it seems to function as expected for those.

It's probably better if you open a separate PR with it.
If it works as expected, you'll then be able to edit this PR.

BowDown097:
> I did test that case with two levels of recursion and it worked fine

Really? 🤔 Linking to an ancestor directory unconditionally causes an infinite loop without the check for visited paths on my machine, and that is what should happen logically. A simple symlink to ``..``, for example, does not cause this for you?

> It's probably better if you open a separate PR with it.

Sure. I could go with that and just close this one (since it solves the problems outlined here anyway).

tbertels:
> > I did test that case with two levels of recursion and it worked fine
> 
> Really? 🤔 Linking to an ancestor directory unconditionally causes an infinite loop without the check for visited paths on my machine, and that is what should happen logically. A simple symlink to `..`, for example, does not cause this for you?

For this specific case indeed the visited paths logic is needed.

> 
> > It's probably better if you open a separate PR with it.
> 
> Sure. I could go with that and just close this one (since it solves the problems outlined here anyway).

👍

BowDown097:
Closed in favor of #615.

PR #615, opened 2025-10-24
Better processing and sorting of [new] playlist items

BowDown097:
This makes sorting of playlist items done in a more "natural" manner, that being case insensitively and with proper handling of numerals.

Helpers::filterUrls has also received these changes:
- Circular symlinks are now checked and skipped.
- Directory traversal is now done separately from top-level files and uses QDirIterator, avoiding recursion and giving slightly better performance.
- Retrieval of normalized file suffixes has been slightly optimized and unified into its own method, used here and elsewhere as applicable.
- Resultant items are now sorted (with the new routine).

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -3,10 +3,12 @@
 #include <QApplication>
 #include <QMouseEvent>
 #include <QAction>
+#include <QDirIterator>
 #include <cmath>
 #include <QRegularExpression>
 #include <QStandardPaths>
 #include "helpers.h"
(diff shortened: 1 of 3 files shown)


cmdrkotori:
LGTM

tbertels:
Works fine for me too, thank you @BowDown097!

tbertels:
We'll probably want to use a QCollator too in the future for `PlaylistWindow::sortPlaylistByLabel`.

PR #641, opened 2025-11-22
Support resizing the video

### LINKED ISSUE #352: Video crop support or Zoom
I built mpc-qt from the source to see how the latest functions work. I used Linux Mint XFCE in a Virtual Machine for this. I tested Pan&Scan. If I am not mistaken currently only the horizontal "stretching" is added to the program capabilities. I tested it on a sample 4:3 video, however when clicking in the menu on the "Increase Pan and Scan" item (or any other Pan&Scan item in fact) nothing happened. The video remained exactly the same [with black bars moving nowhere]. Is it still under implementation?
I think it would be great if the  0.01 value for stretching can be used separately horizontally and vertically and by this a more precise handling can be achieved.
Adding to the Pan&Scan menu "Moving the image left / right / up / down" (separately) with the value 0.05 would further increase the usefulness of Pan&Scan  allowing to position an image to fit the screen even more precise, similarly like in MPC-HC /BE.

Image

Generally the "new" version seems to be more stable and responsive, and I especially like the added subtitle changing option on the toolbar.

tbertels:
Resizes the video without - unlike zoom - preserving the aspect ratio.

No shortcuts are assigned by default as this is likely a less used feature and it would use up the last available numerical shortcuts.

Fixes #352 (Video crop support or Zoom).

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -2159,6 +2159,12 @@ void MainWindow::setVideoTracks(QList<Track> tracks)
     ui->menuPlayVideoZoom->addAction(ui->actionDecreaseZoom);
     ui->menuPlayVideoZoom->addAction(ui->actionIncreaseZoom);
     ui->menuPlayVideoZoom->addAction(ui->actionResetZoom);
+    ui->menuPlayVideo->addMenu(ui->menuPlayVideoResize);
+    ui->menuPlayVideoResize->addAction(ui->actionDecreaseWidth);
+    ui->menuPlayVideoResize->addAction(ui->actionIncreaseWidth);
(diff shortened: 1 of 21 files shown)


tbertels:
That makes me think that the Pan and Scan may not be that much needed anymore now that we have Zoom, which works even with hard coded black bars.
Or we could remove the Decrease and Increase Pan and Scan commands and just keep the Minimum and Maximum ones as it can be useful to easily remove the (not hard coded) black bars by cropping the video.

cmdrkotori:
Merged, though it might need a check for video height/width scale going negative.

PR #658, opened 2025-11-30
Add checks for video resize, fix keyboard shortcut, rename command

tbertels:
Suggested in #641.

Follow-up to 4b24f6ce05fb23c99ad0f1f5929fdbf36a791ace.


cmdrkotori:
The translations need another wash through with cmake.

PR #697, opened 2025-12-18
Reorganize the repository

tbertels:
* packaging: Move appimage, flatpak and nsis to packaging/ directory

> Rename deploy/dist/ to packaging/ to simplify.
> The AppImage folder was in the repository root, so let's move it to packaging
> and change it to lowercase.

* data: Move .desktop and AppStream files to data/

> And rename io.github.mpc_qt.mpc-qt.appdata.xml to
> io.github.mpc_qt.mpc-qt.metainfo.xml as it's the current recommended
> naming.

* scripts: Move script files to scripts/

* src: Move source files to src/

> Along with the other moves, it makes it easier to differentiate between source,
> ressources and other files.
> It also makes the readme easier to get to on GitHub since the desktop browser
> version doesn't yet collapse the file list by default.
> 
> Uses a secondary CMakeLists.txt file.

* res: Move resource files to res/

> Includes http, images and text folders.
> 
> Uses a secondary CMakeLists.txt file.
> 
> settingswindow.ui is updated with the new path to the theme icons in the
> repository. I've manually updated the translations since we're late in
> the release cycle.

See https://github.com/tbertels/mpc-qt for an overview of the new directory structure.


tbertels:
The AppImage workflow should fail because it fetches the master branch but expects the newer paths.
Update: I guess it fails only on forks.

cmdrkotori:
Tested, Works On My Machine

PR #709, opened 2025-12-22
Simplify title bar setting

tbertels:
* Simplify title bar setting
Use a radio button for the "file title" option and rename it.
* mainwindow: Don't set resources in mainwindow.ui
Should permanently fix the problem with qtcreator mentioned in https://github.com/mpc-qt/mpc-qt/commit/30d1d0d4e11d3b2f89583cfbc48fb51529831417.
* mainwindow: Use more descriptive name for title item of information widget
This is the Information panel from View > Information.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1179,8 +1179,8 @@ void MainWindow::updateInfostats()
     bool infoShow = ui->actionViewHideInformation->isChecked();
     bool statShow = ui->actionViewHideStatistics->isChecked();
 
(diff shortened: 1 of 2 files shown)


cmdrkotori:
LGTM. There's a network error in the Flatpak build step, but that's probably on github's end as the file it caught on downloads on my machine.

tbertels:
Yes, it happens regularly with the Flatpak workflow.
The same thing happened with the AppImage worflow, but I've added retries.

PR #718, opened 2025-12-27
Use constexpr for Logger prefix names and add ui creation logging

tbertels:
* Use static constexpr instead of constexpr or const char
* Use constexpr for Logger prefix names
Makes it easier to add more logging.
* Add ui creation logging


tbertels:
For mpvwidget, the constexpr are in the class declarations instead of the top of the .cpp file because three different names are used.

PR #722, opened 2025-12-29
Always use palette detection for fallback theme icons and implement newer method

tbertels:
* helpers: Implement newer method to get color scheme for fallback icons

> As an aside, it fixes color scheme detection in the Windows CI tests.

* settings: Always use palette detection for fallback theme icons https://github.com/mpc-qt/mpc-qt/commit/

> b9d3d88bf27a8701c7057512d6fb13993235fc16 introduced a regression where the fallback icons palette selector is disabled until the theme QComboBox is changed from and back to fallback icons.
> 
> Since https://github.com/mpc-qt/mpc-qt/commit/e865fb7cdb927b5de779557a622a66bd347fcdf5, the fallback icons palette selector was just there in case the detection didn't work.
> So let's just remove that option instead of fixing it.
> 
> It also increases the vertical space available to the Widget Color QScrollArea.

### DIFF
--- a/.github/actions/data/settings.json
+++ b/.github/actions/data/settings.json
@@ -132,7 +132,6 @@
     "hwdecVp9": true,
     "hwdecWmv3": true,
     "interfaceIconsCustomFolder": "",
-    "interfaceIconsInbuilt": 0,
     "interfaceIconsTheme": 0,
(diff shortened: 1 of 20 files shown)


cmdrkotori:
Definitely an improvement.

PR #765, opened 2026-01-16
ci: Add two additional attempts to Flatpak workflow

tbertels:
It sometimes fails retrieving the source of libXpresent on the freedesktop website.


cmdrkotori:
`meson.build:1:0: ERROR: Meson version is 0.61.2 but project requires >=0.63` in the appimage action.

Not sure where we're going to find a more recent version of meson other than compiling that from source too.

tbertels:
I've found it there: https://www.ubuntuupdates.org/package/kubuntu-ppa_backports_extra/jammy/main/base/meson
I was expecting it to be 0.63.2-1, but it's 1.7.2.

PR #788, opened 2026-01-31
Use latest Qt LTS for the CI and the AppImage

tbertels:
* Use latest Qt LTS for the CI and the AppImage

> Thanks to https://github.com/jurplel/install-qt-action we can now use
> the latest LTS version of Qt (currently 6.8) on Ubuntu 22.04 (instead of 6.2).
> Note that Ubuntu 24.04 (to which we'll switch in 2027) uses Qt 6.4,
> so we'd have to wait until 2029 and Ubuntu 26.04 to use Qt >=6.5.
> 
> AppImage tested on Ubuntu 22.04, it works fine.
> It fixes the background for drawnslider, and the languages names in
> tracks submenus (6.5 needed).
> 
> Link: https://packages.ubuntu.com/search?keywords=qt6-base-dev
* Remove checks for Qt 5

> It doesn't build on Qt 5 anyway.
> 
> Removing support for versions older than 6.8 isn't urgent as it wouldn't
> be that useful even in the CMake file.
> Support isn't guaranteed in the future since we'll build in the CI with
> Qt 6.8 from now on, but it can be fixed if someone needs it.

### DIFF
--- a/src/favoriteswindow.cpp
+++ b/src/favoriteswindow.cpp
@@ -111,11 +111,7 @@ FavoritesItem::FavoritesItem(QListWidget *owner, const TrackInfo &t)
 {
     track_ = t;
     Qt::ItemFlags f = flags();
-#if QT_VERSION >= 0x050700
     f.setFlag(Qt::ItemIsEditable);
-#else
-    f |= Qt::ItemIsEditable;
-#endif
     this->setFlags(f);
 }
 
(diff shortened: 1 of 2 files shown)


cmdrkotori:
It also should fully enable hwdec on wayland, as we can pass wl_display to mpv now.

tbertels:
> It also should fully enable hwdec on wayland, as we can pass wl_display to mpv now.

Yes, it does work with the native build binary.
The AppImage doesn't yet have support for wayland mode though.

PR #794, opened 2026-02-01
Support changing status time format by click, add remaining and percentage modes

### LINKED ISSUE #463: Feature request: high precision timestamp in right click option
Image

Image

Image

tbertels:
* Support changing status time format by click and add remaining and percentage modes

> Remaining mode shows the how much time is remaining; percentage mode shows completion.
> A left click switches between normal and remaining time mode.
> The context menu allows enabling "Remaining time", "High precision" and "Show percentage".
> 
> Uses only one StatusTime widget. This avoids what seems like a Qt bug
> where the widgets would quickly wobble when their widths change.
> 
> Unlike MPC-HC, moving the player from that position still works.
> 
> Fixes https://github.com/mpc-qt/mpc-qt/issues/463 (Feature request: high precision timestamp in right click
> option).
* Always use playback time short mode in tooltip and chapters

> The status time widget is where the high precision mode makes most sense.
> 
> Follows the  behavior of MPC-HC.
> 
> Follow-up to and partially reverts
> https://github.com/mpc-qt/mpc-qt/commit/b06e81dcb7e6312abfcdd083a08fe375f50d1b82 before which the seekbar
> tooltip and chapters never used the playback time short mode.
* Don't show StatusTime when player is stopped

> Avoids showing the status time when nothing useful is available. Mirrors
> the behavior of Drawnslider.
> 
> We may want to mirror the behavior of MPC-HC in the future: hide the
> status time only when no file is loaded and show otherwise its total
> length (and keep it loaded). MPC-HC seems to pause the file and move the
> position back to 0, but that keeps the last image visible.

### DIFF
--- a/.github/actions/data/settings.json
+++ b/.github/actions/data/settings.json
@@ -415,7 +415,6 @@
     "tweaksPreferWayland": false,
     "tweaksSeekFramedrop": false,
     "tweaksShowChapterMarks": true,
-    "tweaksTimeShort": true,
     "tweaksTimeTooltip": false,
     "tweaksTimeTooltipLocation": 0,
     "tweaksVideoPreview": false,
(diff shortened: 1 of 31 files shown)


tbertels:
Here's what it looks like: 

image

PR #806, opened 2026-02-09
Adding StartupWMClass to the .desktop file

Kyuyrii:
This is to avoid problems with DEs/compositors using the wrong icon.


tbertels:
Thanks for the PR!
I'm guessing this is for Ubuntu and the Snap version? With which DE have you noticed that problem?

Kyuyrii:
> With which DE have you noticed that problem?

I'm currently using KDE, but WMClass is used by other DEs/compositors.

I've noticed problems with WMClass in several apps; apparently, WMClass can be different in Xorg and Wayland.

The problems can include, for example, the Wayland icon being used instead of the app's icon, or even the pinned app being considered a different app, for example:

Captura de imagem_20260209_194314


So, since some apps don't have StartupWMClass or use the Xorg one, it can cause some problems in Wayland DEs/compositors.

tbertels:
> I'm currently using KDE, but WMClass is used by other DEs/compositors.

It seems to be an issue specific to Kubuntu 25.10 with the Snap version. And it's even worse when MPC-QT is closed while being pinned to the taskbar:

image

Adding `StartupWMClass=io.github.mpc_qt.mpc-qt` does fix it.
I've tested with Ubuntu 22.04 and Kubuntu 24.10 but I haven't noticed that issue with the Snap version.

What I've noticed on Kubuntu when MPC-QT (Snap version) is pinned is the icon disappearing when closing it and reappearing a few seconds later.
Adding `StartupWMClass=io.github.mpc_qt.mpc-qt` didn't fix it and `StartupWMClass=mpc_qt.mpc-qt` did show a generic Wayland icon instead of the normal one.

PR #807, opened 2026-02-11
Improve "Interface" settings page

tbertels:
* settingswindow: Reformat .ui file with Qt Designer
* settingswindow: Move Interface "Video" and "Info" colors to widgets scrollbox

> To increase the available space for the Interface colors settings.
> 
> And rename "Window Color" to "Others".
> 
> The "Use custom colors" checkbox now applies to them too.

* settingswindow: Rename "Info" and "Video" color settings

> Rename "Info [Background|Foreground]" to "Status bar [background|foreground]".
> Rename "Video" to "Video background".

* settingswindow: Rename "Use custom theme" to "Use custom colors"

> To better separate this setting from the icons setting.

* settingswindow: Adjust wording of icons-related settings

> Rename "Theme" to "Source" and "Built-in fallback" to "Built-in" to reduce confusion.
> Change the message about custom icons naming.

* paletteeditor: Connect "Button & Window" button to its actual function

> It was connected to the same function as "Button".

* paletteeditor: Make unused colors less visible and rename "Generate"

> The buttons next to "Generate" are used to generate a palette from
> either the "Button" color or both the Button" and "Window" colors.
> So "Generate palette from:" is a little more easier to understand.

* Clean up auto icons-related code

> Since https://github.com/mpc-qt/mpc-qt/commit/deb6c8cb8e5fad309201d47dc4f13f65c2e480b5 we always detect the palette when using the built-in icons, so that code isn't needed.

* helpers: Rename fallbackFolder to builtinFolder

> 9fe30523b7c5476b009b205f28413114131164d7 renamed the option, which isn't actually a fallback since https://github.com/mpc-qt/mpc-qt/commit/d9bbda574969a1ce5d47d6b3e817ee9c8fed47c2.

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -744,12 +744,12 @@ QIcon IconThemer::fetchIcon(const QString &name)
     }
     QIcon icon;
     if (folderMode_ != SystemFolder) {
-        icon = QIcon(fallbackFolder_ + name + ".svg");
+        icon = QIcon(builtinFolder + name + ".svg");
     }
     else {
-        icon =  QIcon::fromTheme(name, QIcon(fallbackFolder_ + name + ".svg"));
+        icon =  QIcon::fromTheme(name, QIcon(builtinFolder + name + ".svg"));
     }
-    if (fallbackFolder_ == whiteIconsPath) {
+    if (builtinFolder == whiteIconsPath) {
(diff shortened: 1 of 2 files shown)


tbertels:
Here's before:

o1

o2


And after:

n1

n2

PR #812, opened 2026-02-14
Show message when ambiguous shortcut is used

tbertels:
This lets the user know what is going on when Qt filters out a key press because it is assigned to more than one command.

The pointer to `mainWindow` is needed to center the message box relative to it.

I'm not too happy about showing that message from the Logger class, but that's where the filtering happens as `qInstallMessageHandler` seems to be the only way to catch it.

### DIFF
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -1,5 +1,6 @@
 #include <cstdio>
 #include <cstdlib>
+#include <QMessageBox>
 #include <unistd.h>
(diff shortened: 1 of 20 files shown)


cmdrkotori:
>I'm not too happy about showing that message from the Logger class, but that's where the filtering happens as qInstallMessageHandler seems to be the only way to catch it.

That's fine.

Issue #814, opened 2026-02-17
[Feature request] after playback shutdown


mynameistakenhowcanthatb:
like in K-Lite Codec . after a playlist is finish pc turns off . its grate for people that put a playlist on and fall asleep to tv/movies

tbertels:
Have you tried this:

<img width="580" height="841" alt="Image" src="https://github.com/user-attachments/assets/7a475673-e649-440a-9bb8-c54f2fee5b8f" />

mynameistakenhowcanthatb:
my one dont seem to have that. sorry still new to this app 

<img width="495" height="394" alt="Image" src="https://github.com/user-attachments/assets/67adb824-5f52-4492-895e-8257f3edae22" />

tbertels:
The Flatpak version needs access to `org.freedesktop.login1`.

There's also the fact that Hibernate, Suspend and PowerOff don't currently work, but that will be fixed by #816.

mynameistakenhowcanthatb:
that explains it . lol was thinking i was going crazy looking over all the options menu . thank you

tbertels:
Fixed in Flatpak version by https://github.com/flathub/io.github.mpc_qt.mpc-qt/pull/85 and https://github.com/flathub/io.github.mpc_qt.mpc-qt/pull/87

PR #816, opened 2026-02-18
screensaver: Add missing interactive argument

tbertels:
org.freedesktop.login1.Manager Hibernate, Suspend and PowerOff require an interactive argument. It is used to control whether polkit should interactively ask the user for authentication credentials if required. These methods don't work without it.

Link: https://www.freedesktop.org/software/systemd/man/org.freedesktop.login1.html

#814 ("[Feature request] after playback shutdown")

### DIFF
--- a/src/platform/screensaver_unix.cpp
+++ b/src/platform/screensaver_unix.cpp
@@ -99,7 +99,7 @@ void ScreenSaverUnix::hibernateSystem()
     if (!isInhibiting)
         return;
 
-    QDBusPendingCall async = dbusLogin.asyncCall("Hibernate");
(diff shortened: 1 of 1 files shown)


tbertels:
`logOff()` seems broken on my system, it doesn't close the session cleanly.

tbertels:
> `logOff()` seems broken on my system, it doesn't close the session cleanly.

It works fine on Ubuntu 22.04 and Mint 22.2 so it's either a KDE bug or a problem on my system.

Same problem with Kubuntu 24.10, so it's probably a KDE bug.

PR #839, opened 2026-03-06
Reduce the size of buttons

### LINKED ISSUE #820: Smaller buttons and a little bit less space for control and seek bar
Image

In MPC-HC, (left side) buttons are small and look cute. Sometimes we want to see videos with the maximum viewing area without going full screen. Removing the menu and status bar helps in that case, but a little bit more space can be saved in the buttons and bar. Please look into this. 

Thanks.

tbertels:
* Reduce buttons size to 12px instead of default 16px

> And reduce controls vertical lines height to 20px.
> 
> Follow-up to https://github.com/mpc-qt/mpc-qt/commit/b7ec5c17cf1f68a9601651ad5a5fe12e5b59f2a8 which made the icons use all the available space in the buttons.
> 
> Fixes https://github.com/mpc-qt/mpc-qt/issues/820 ("[Suggestion] Smaller buttons and a little bit less space
> for control and seek bar").

* mainwindow: Reduce buttons spacing to 2px instead of default 6px

* mainwindow: Remove control bar top margin

> It improves the vertical centering of the controls between the the seek bar and the status bar.
> 
> It was only wasting useful space for the video, because the margin between the buttons and the seek bar is still larger than the margin between the buttons and the status bar.
> 
> Fixes https://github.com/mpc-qt/mpc-qt/issues/820 ("[Suggestion] Smaller buttons and a little bit less space for control and seek bar").

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -130,7 +130,7 @@
               <number>6</number>
              </property>
              <property name="topMargin">
-              <number>3</number>
+              <number>0</number>
              </property>
              <property name="rightMargin">
               <number>6</number>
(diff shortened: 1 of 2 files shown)


tbertels:
Here's before:

before

And after:

after

tbertels:
As a comparison, here's what it looked like before we increased the icons size:

418664120-ae8db997-dcfa-4a35-b768-cba56b96d35d

PR #851, opened 2026-03-20
helpers: Guard against malformed input in date/time parsers

tbertels:
`GoToWindow::on_goToTime_textChanged` caused a segfault in `Helpers::fromDateFormat` when pasting a number after selecting the content.


tbertels:
The msys2 error is because of https://github.com/ImageMagick/ImageMagick/issues/8630.

PR #886, opened 2026-04-05
cmakelists: Use GLOB to add all translations (including ko and forgotten pl one)

tbertels:
We forgot to add the Polish translation so that the translators couldn't test it.
Since they'll now be added automatically, this won't happen again.


tbertels:
@radzik136 and @Kieubasiarz: sorry for the delay.

PR #893, opened 2026-04-10
Use custom slider widget for audio balance and color controls settings

tbertels:
Replace scrollbars with a new custom slider widget based on QSlider.

QSlider has a highlight in its groove between the minimum position and the current one, so it cannot be used when the neutral position is at the center.
Unlike QSlider, this custom widget can be styled in the settings even when not using the Fusion style. It's also more visible with dark themes than QSlider.

Tested with various themes including Fusion (both dark and light).


tbertels:
Here are comparisons between QSlider, scrollbars and the new "PlainSlider" widget with the Breeze and Fusion themes in dark and light colors:

dark plainslider
light plainslider
dark fusion plainslider corr
light fusion plainslider corr

PR #896, opened 2026-04-14
Use lambda expressions for QMetaObject::invokeMethod calls

tbertels:
Avoids runtime method lookup and improves robustness and readability.


cmdrkotori:
Do these still execute on the remote thread?

cmdrkotori:
Forget that, I just read the docs. Yes it does!

PR #903, opened 2026-04-18
drawnslider: Fix background image height calculation

### LINKED ISSUE #891: Seekbar causes video to stutter slightly
After getting a 4k monitor I noticed that whenever the OSD (specifically the seek bar) is present at the bottom of the screen and I'm in fullscreen, the video playback will stutter a little bit, not be as smooth as before. It's most noticeable in usually smooth video (e.g. 30+ fps) but it doesn't seem to be based around the source video, it happens to anything. I never noticed this with my older 1080p monitor so uncertain if the problem is at my end or if it was always a problem but only just starts to manifest at 4k. It's very subtle however so it might not be noticeable to many people, but it's there for me at least.

Running MPC-QT dev build (9/4/26 at 9:22 am), Flatpak version on Kubuntu 25.10 (Wayland). Using NVIDIA card with HW acceleration.

EDIT: Clarifying that the OSD that shows text, timer and whatnot is fine. No dropped frames according to libmpv OSD. When I say OSD, I specifically mean the **seekbar**. Probably should have said that in the title, whoops!

tbertels:
The background image was allocated using width() for both dimensions, resulting in a square buffer (width × width) instead of (width × height).

Estimated impact on RAM usage in maximized or fullscreen mode:

With a 1920×1080 layout and a 1905x35 slider:
- Before: 14.8 MB (1905×1905×4)
- After: 0.3 MB (1905×35×4)

With a 4K screen:
- Before: 55.6 MB (3818×3818×4)
- After: 0.7 MB (3818×49×4)

Measured impact at 1920×1080:
- ~50% reduction in seekbar CPU usage when comparing with vs without seekbar

Fixes: 0fff34c37d1b18d571fab3be713b4870814e7c8e ("slider: use bitmaps to display sliders")

This may also help with #891 ("Seekbar causes video to stutter slightly").

### DIFF
--- a/src/widgets/drawnslider.cpp
+++ b/src/widgets/drawnslider.cpp
@@ -341,7 +341,7 @@ void MediaSlider::makeBackground()
     Logger::log(logModule, "MediaSlider::makeBackground");
     qreal pr = devicePixelRatioF();
     int pw = width() * pr;
-    int ph = width() * pr;
+    int ph = height() * pr;
     backgroundPic = QImage(pw, ph, QImage::Format_RGBA8888);
     backgroundPic.fill(Qt::transparent);
     QPainter p(&backgroundPic);


tbertels:
@cmdrkotori I'd be curious to know which numbers you're seeing with your 4K screen at 175% scaling.

cmdrkotori:
With this patch:
Running a 1080p video at 100% gives me a background dimension of 1898x48.
Running a 1080p video in a maximized window gives a background dimension of 3818x49. (I don't know where the extra pixel comes from.)
In any case this a clear improvement over the square allocation.

tbertels:
I meant about the RAM usage as the number looks quite extreme.

cmdrkotori:
With nothing open, default window size, before patch: 362Mb
With nothing open, default window size, after patch: 347MB

tbertels:
These numbers are both much closer and much bigger than I would expect.
Here it shows 141MB vs 114MB maximized with nothing opened.

Is it with the window maximized? And are they the default "Memory" value shown by plasma-systemmonitor?

cmdrkotori:
Right, I was at default window size and using htop, which I seem to have misread.  I learned something today. With a maximized mpc-qt window, plasma-systemsettings shows 239.2Mb (pre-patch) and 171Mb (post-patch).  That is somewhat more believable.

tbertels:
Thanks, I've adjusted the numbers, because they were incorrectly combining the scaling and the original resolution.
The ~55 MB number is more in line with what you're getting.

PR #940, opened 2026-05-11
Don't play first frame when resuming position

tbertels:
It's faster than watching the seekable property, which avoids playing the first frame.

Tested with streams and yt-dlp.

Follow-up to f02b1a4d33629ad1a44290689b2913115a550a2c ("helpers,main,manager,mpvwidget: implement start times").


tbertels:
Looks like we've found a bug in the way git (not just GitHub, gitk too) shows
```
+void MpvObject::setStartTime(double position)
+{
+    setMpvPropertyVariant("start", QString::number(position));
+}
+
 void MpvObject::setTime(double position)
 {
     setMpvPropertyVariant("time-pos", position);
 }
 
-void MpvObject::setTimeSync(double position)
-{
-    ctrl->command(QVariantList() << "seek" << position << "absolute");
-}
-
```

cmdrkotori:
Yeah, concise ways to calculate a diff can be tricky.

PR #941, opened 2026-05-12
widgets: jump to next chapter on middle-click in progress bar

Thoxy67:
Middle-clicking the position slider seeks to the first chapter whose start time is after the click position. 
No-op when the file has no chapters or the click is past the last one.

Mirrors Haruna's `HProgressBar.qml` behavior (one of my favorite linux video player).

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -792,6 +792,8 @@ void MainWindow::setupPositionSlider()
             this, &MainWindow::position_hoverValue);
     connect(positionSlider_, &MediaSlider::hoverEnd,
             this, &MainWindow::position_hoverEnd);
+    connect(positionSlider_, &MediaSlider::middleClicked,
+            this, [this](double t) { emit timeSelected(t); });
 }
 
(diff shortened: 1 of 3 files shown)


cmdrkotori:
Excellent idea! Thanks!  I'll wait for the checks to happen, then merge this when they're done.

tbertels:
Nice, thanks for the PR!

tbertels:
I'm wondering if it wouldn't be better to jump to the start of the current chapter because:
- we show the title of the current chapter in the tooltip / video preview
- it makes it possible to seek to the start of the first chapter (seeking to the end of the last chapter isn't very useful)

tbertels:
> I'm wondering if it wouldn't be better to jump to the start of the current chapter because:
> 
>     * we show the title of the current chapter in the tooltip / video preview
> 
>     * it makes it possible to seek to the start of the first chapter (seeking to the end of the last chapter isn't very useful)

@Thoxy67: Do you think it would be a bad idea?

Issue #935, opened 2026-05-09
Make mouse wheel (volume setting) more responsive


subcor:
CachyOS x86_64 KDE Plasma 6.6.4 Wayland
Built on 2026. 05. 04. at 22:03 Flatpak

I don't know if I should file this as a bug report or feature request for improvement:
The volume slider's sensitivity can be inconsistant when setting it with mouse wheel, kinda ignoring a lot of the scrolling input.
This seems to be related to the framerate of the video, since it's much smoother if no video or a 60FPS video is being played than a video around 24FPS.

tbertels:
I haven't been able to reproduce this.
With the Flatpak and the native build, every scroll step results in a volume change.

Have you checked your CPU usage?
Do you use a sched_ext?

subcor:
What I'm talking about is not steps, but longer scrolls.
For example try and compare how quickly can you get the volume from 100% to 0% with no file loaded, and compare it while a 24FPS video being played.
The CPU usage is not that high, and even then a 1080p H265 24FPS video with lower CPU usage is less responsive than a 2160p VP9 60FPS video.

No idea what sched_ext is, but I typed this in console:
grep ext /proc/self/sched
and it returned with:
ext.enabled : 0

tbertels:
Does it happen only in fullscreen? And does your screen have variable refresh rate?

This may be something that will get better in Plasma 6.7 with https://www.phoronix.com/news/KDE-Plasma-Faster-WL-SHM

subcor:
I initially tested it in window but happens in full screen too. My screen has fixed refresh rate.
I just tried it on a different PC with Linux Mint 22.3 Cinnamon 6.6.7 (X11) and it behaves the same.
edit: I also tried mpv and vlc but those don't seem to be affected

subcor:
Hm, if I enable "Prefer Wayland over Xwayland" it improves the responsiveness a lot, pretty much fixes it. But of course this option is not available on X11 (like Mint).

tbertels:
Wayland is coming to Mint too, here's the current state of Wayland support for Cinnamon: https://trello.com/b/HHs01Pab/cinnamon-wayland

subcor:
Yeah, I guess with time this issue will solve by itself, and since the "severity" of this issue is very low anyway it's probably not worth spending time on. Btw, is there a reason to use xwayland with mpc-qt?

tbertels:
> Btw, is there a reason to use xwayland with mpc-qt?

[...]

PR #942, opened 2026-05-12
drawnslider: cache background image, only redraw on state change

### LINKED ISSUE #891: Seekbar causes video to stutter slightly

Thoxy67:
The seekbar was re-rasterizing the groove + chapter ticks + A-B loop highlight every paint, even though those only change on resize, palette change, or new chapters/loop points. During playback that's the same image being redrawn dozens of times per second.

Now caches the background QImage and only regenerates it when something affecting it actually changed. setMaximum/setMinimum also check whether the value really differs before invalidating, since MainWindow::setTime calls setMaximum(length) on every time-pos
  tick with the same value — without that check the cache would still flush every frame.

Also drops the per-paint Logger::log lines from both makeBackground overrides — they were firing at frame rate and adding log noise for no real diagnostic value.

### DIFF
--- a/src/widgets/drawnslider.cpp
+++ b/src/widgets/drawnslider.cpp
@@ -1,8 +1,8 @@
 // A simple(?) set of sliders related to controlling multimedia players.
 #include <QPainter>
 #include <cmath>
-#include "drawnslider.h"
 #include "logger.h"
+#include "drawnslider.h"
 
 static constexpr char logModule[] =  "drawnslider";
 
@@ -38,7 +38,10 @@ void DrawnSlider::setValue(double v)
 
 void DrawnSlider::setMaximum(double v)
 {
-    vMaximum = v;
+    if (vMaximum != v) {
+        vMaximum = v;
+        redrawBackground = true;
+    }
     if (vValue > v)
         setValue(v);
     else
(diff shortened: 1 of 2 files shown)


tbertels:
Thanks for fixing this! It's hopefully the last piece needed to fix #891.
It doesn't affect #935 but I'm curious to see if Plasma 6.7 won't fix it.

> Also drops the per-paint Logger::log lines from both makeBackground overrides — they were firing at frame rate and adding log noise for no real diagnostic value.

Yes, they were annoying. I was just keeping them as a reminder that makeBackground was being called too often.

Could you just keep the `logModule` line? Its purpose is to avoid having to manually enter the name when we do use logging during development.

Thoxy67:
Restored the logModule line (and the logger.h include) so it's ready to use whenever a log call needs to go in. Glad to hear this might wrap up #891 🤞

PR #943, opened 2026-05-12
drawnstatus: skip repaint when displayed time hasn't changed

Thoxy67:
StatusTime::setTime is called on every mpv time-pos update (frame rate), but the formatted text is second-precision. So most ticks rebuild the exact same "01:23 / 45:67" string and trigger a paint with identical pixels.

Caches the last built string and bails before update() when it matches. The format-toggle setters (short / remaining / percent) clear the cache so switching mode still repaints immediately.


tbertels:
Thanks!

PR #944, opened 2026-05-12
settings: add Replay Gain controls to Audio renderer page

### LINKED ISSUE #736: Equivalent option to normailze audio volume / volume normalization?
Sorry to bother you,
in MPC-HC, under Options - Internal Filters - Audio Switcher, there is a 'Normalize' checkbox along with some other settings like "regain" and "boost"

This is useful,
when the audio volume is too low, I can't hear what people are saying, and this feature can make the dialogue louder;  
when the volume is too high, for example with some horror sound effects or background music, this feature can lower the volume.  

Is there an option similar to this?

Thoxy67:
mpv has Replay Gain support built in (replaygain={no,track,album}, replaygain-preamp, replaygain-clip, replaygain-fallback), but there was no UI to actually turn it on. People with Replay-Gain-tagged music libraries (foobar2000 / Picard / rsgain) had to edit input.conf or pass mpv options manually.

Adds a "Replay Gain" group on the audio settings page with four controls: Mode (No / Track / Album), Pre-amp (dB), Fallback (dB, applied when a file has no Replay Gain tag), and Prevent clipping. Wired through the existing audio-option pattern, the combo maps to mpv's enum strings via indexedValueToText, and sendSignals() emits the four option() signals that MpvObject already consumes.

### DIFF
--- a/src/settingswindow.cpp
+++ b/src/settingswindow.cpp
@@ -164,6 +164,7 @@ QHash<QString, QStringList> SettingMap::indexedValueToText = {
     {"audioChannels", {"auto-safe", "auto", "stereo", "mono"}},
     {"audioRenderer", {"pulse", "alsa", "oss", "null"}},
     {"audioAutoloadMatch", { "exact", "fuzzy", "all" }},
+    {"replayGainMode", {"no", "track", "album"}},
     {"framedroppingMode", {"no", "vo", "decoder", "decoder+vo"}},
     {"framedroppingDecoderMode", {"none", "default", "nonref", "bidir",\
                                   "nonkey", "all"}},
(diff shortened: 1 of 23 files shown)


tbertels:
You forgot to update translations with `cmake --build build -t update_translations`.
A simple way to avoid forgetting it is to use
 `cmake -G Ninja -B build ; cmake --build build -t update_translations ; cmake --build build`
to build.

About the Replay Gain controls, to limit the increase in vertical height of the Options window, I think it would better to place them on the right of the "Audio open settings", where the audio balance is currently. And the audio balance could be moved for instance below the "Audio open settings" (though not ideal):

image

Thoxy67:
I have updated translations as needed and also changed the layout to put the audio balance on one column so it doesn't look weird anymore 😂

image

tbertels:
Thanks!

When you push a commit to modify a previous one, you can actually just amend the previous one, force push and describe the changes in a comment. That way it's easier to merge the PR as we don't need to rebase manually.
When a PR includes multiple commits (for related changes), changes to individual commits can be done with fixup commits and a force push at the end.

PR #945, opened 2026-05-12
widgets: add status time hover tooltip and double-click goto

Thoxy67:
Last PR of the day, wrapping up a few small things I felt were missing compared to other players I've been using.

Three additions on the time display:

  - A "Show milliseconds in time display" checkbox under Tweaks, so it can be set from settings instead of right-clicking the time.
  - Double-click on the time in the status bar opens the Go To Time dialog.
  - When remaining-time mode is on, the display shows -remaining / played instead of -remaining / total. Normal mode is unchanged (played / total).

Also wanted to say really awesome project. I've already switched from Haruna to mpc-qt as my main player on CachyOS.

I hope you enjoyed my small PR so far and would be happy to help in the future if I see something relevant to fix or implement.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -398,6 +398,8 @@ void MainWindow::changeEvent(QEvent *event)
             tooltip->updatePalette();
         if (videoPreview)
             videoPreview->updatePalette();
+        if (statusTime)
+            statusTime->updatePalette();
     }
     else if (event->type() == QEvent::WindowStateChange && isMaximized())
         emit windowMaximized();
@@ -509,11 +511,7 @@ void MainWindow::mouseReleaseEvent(QMouseEvent *event)
 {
     QPoint pos = event->globalPosition().toPoint();
     bool ok = mpvw ? insideWidget(pos, mpvw) : false;
-    if (mousePressedInBottomArea && insideWidget(pos, statusTime)) {
-        if (event->button() == Qt::MouseButton::LeftButton)
-            setTimeRemainingMode(!timeRemainingMode);
-    }
-    else {
+    if (!(mousePressedInBottomArea && insideWidget(pos, statusTime))) {
         if (ok && mouseStateEvent(MouseState::fromMouseEvent(event, MouseState::MouseUp)))
             event->accept();
         else
@@ -882,6 +880,8 @@ void MainWindow::setupStatus()
     ui->statusbarLayout->insertWidget(2, statusTime);
     connect(statusTime, &StatusTime::customContextMenuRequested,
             this, &MainWindow::statusTime_customContextMenuRequested);
+    connect(statusTime, &StatusTime::doubleClicked,
+            ui->actionNavigateGoto, &QAction::trigger);
 }
 
 void MainWindow::setupSizing()
--- a/src/widgets/drawnstatus.cpp
+++ b/src/widgets/drawnstatus.cpp
@@ -1,5 +1,7 @@
 #include <QPainter>
+#include <QMouseEvent>
 #include "helpers.h"
+#include "tooltip.h"
 #include "drawnstatus.h"
 
 static constexpr char logModule[] =  "drawnstatus";
@@ -8,6 +10,7 @@ StatusTime::StatusTime(QWidget *parent) : QWidget(parent)
 {
     setTime(0, 0);
     setMinimumSize(minimumSizeHint());
+    setAttribute(Qt::WA_Hover);
 }
 
 QSize StatusTime::minimumSizeHint() const
@@ -80,6 +83,10 @@ void StatusTime::updateText()
                                                                 'f',
                                                                 1)));
     }
+
+    if (hovered)
+        updateHoverTooltip();
+
     if (drawnText == lastDrawnText)
         return;
     lastDrawnText = drawnText;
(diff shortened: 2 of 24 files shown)


tbertels:
> Last PR of the day, wrapping up a few small things I felt were missing compared to other players I've been using.
> 
> Three additions on the time display:
> 
>   * A "Show milliseconds in time display" checkbox under Tweaks, so it can be set from settings instead of right-clicking the time.

This re-adds (the opposite of) the setting removed by #794, with clearer phrasing.
It was removed because like for MPC-HC, the context menu seemed more user-friendly, and the fewer settings in the settings window, the better.
Did you actually find it not that straightforward?
Note that the widget setting isn't forwarded to the settings window.

>    * Double-click on the time in the status bar opens the Go To Time dialog.

👍

>    * When remaining-time mode is on, the display shows -remaining / played instead of -remaining / total. Normal mode is unchanged (played / total).

This isn't how it's shown in MPC-HC or YouTube. Is there a media player which shows it like that?

> Also wanted to say really awesome project. I've already switched from Haruna to mpc-qt as my main player on CachyOS.

Thanks! ♥️

> I hope you enjoyed my small PR so far and would be happy to help in the future if I see something relevant to fix or implement.

Yes, we did! 🥳 Thanks again!

Thoxy67:
Thanks for the detailed feedback! Quick context, a lot of these little additions came from habits I picked up using VLC for years.

It's getting late and I've got work tomorrow, so I'll properly look at this and push updates when I'm back from work, but here's where my head's at:

- **Milliseconds checkbox**: I added the checkbox alongside the context-menu option mainly for discoverability (the right-click isn't very obvious unless you know to look for it), but I take the point about keeping settings lean per #794. Happy to drop the checkbox and rely on the context menu, that also sidesteps the sync gap you mentioned.

- **"-remaining / played"**: My thinking was that I already know the total duration when I start a video, so showing how much I've already watched felt more useful as the second value. But you're right that no major player does it that way, so I'll revert it to **"-remaining / total"**. Alternatively, I could replace the dual-display with a tooltip on hover (like Haruna does) to show the remaining time without changing the main display at all, let me know if you'd prefer that direction instead.

**Haruna :**
image

**VLC :**
image

More tomorrow!

tbertels:
- remaining / played: the tooltip to show the remaining time is a great idea!
- Milliseconds checkbox: we could also show a message like "Right-click to change time settings" on the second line (with "\n") of the tooltip until the user has right-clicked at least once (saved in the profile settings). Though right-clicking is the established way to show a context menu, and the pointer icon does hint to more features.

For the translations conflicts, the easiest way is to amend the commit to revert the .ts files "edits", pull the master branch and then rebuild.

About the discoverability of features, we could make short screencasts to advertise the different features. KDE Spectacle does make that easy. We could then host them both on the website and elsewhere like YouTube (that's also a good way to advertise MPC-QT itself).

Thoxy67:
Pushed the update! Thanks again for taking the time on the review, the tooltip idea worked out really nicely.

Quick rundown of what changed since the last push.

Dropped the milliseconds checkbox entirely. The right click context menu is the way to toggle it, like you suggested per #794.

Reverted "-remaining / played" back to "-remaining / total" so it matches MPC-HC and YouTube convention.

Added the hover tooltip on the status bar time, Haruna style. It updates in real time as playback progresses, positioned above the time, and shows the complementary value depending on the current display mode. So when you're in normal mode it shows the remaining
time, and when you're already in remaining mode it shows how much you've already played, that way you always get the info you can't see in the main display.

Normal mode, tooltip shows Remaining:
image

With high precision enabled (milliseconds shown):
image

Remaining mode active, tooltip shows Played:
image

Also removed the left click toggle on the time. The right click menu's "Remaining time" entry stays as the toggle, same UX as the "High precision" one right below it.

Kept the double click on the time for the Go To dialog.

For the translations I followed your workflow, reverted the .ts edits, pulled master and rebuilt. Only the two new strings (Played, Remaining) remain in the diff, translated to French.

tbertels:
Nice!

About the hover tooltip, there are two issues:
- it doesn't show anymore if we open Options, or the context menu after moving out of it until we move the window
- it doesn't follow the custom colors set in the Interface settings

The first one can be fixed by parenting the tooltip to StatusTime, though it changes its colors.

The best way to fix both is to use the Tooltip class from tooltip.cpp.
The updatePalette() method of Tooltip can be called from MainWindow::changeEvent through a StatusTime updatePalette() method to follow the palette changes.

By the way, you don't need to manually delete hoverTooltip if you parent it to `this` ( with `hoverTooltip = new Tooltip(this)`). See https://doc.qt.io/qt-6/objecttrees.html.

setMouseTracking isn't needed either with Qt::WA_Hover.

Thoxy67:
Pushed the fix! Thanks for the pointers, the Tooltip class made this much cleaner.

Rundown of what changed:

Swapped the standalone QLabel for the existing Tooltip class from tooltip.cpp. That fixes both issues at once: the tooltip keeps showing after opening Options or the context menu, and it now follows the custom palette colors from Interface settings via Tooltip::updatePalette().

Added StatusTime::updatePalette() and wired it into MainWindow::changeEvent next to the existing tooltip and videoPreview calls, so palette changes propagate through.

Dropped the manual delete hoverTooltip in the destructor and removed setMouseTracking(true) since Qt::WA_Hover already covers the hover events. The destructor itself is gone since Qt's object tree handles cleanup now.

One small judgment call: instead of new Tooltip(this), I parent it to window(). The reason is that Tooltip parents its internal QLabel to whatever you pass in, and child widgets get clipped to their parent's geometry. StatusTime is only status-bar sized, so the tooltip would render clipped. Using window() gives MainWindow as the parent (same Qt auto-cleanup, wide enough drawing surface). The tooltip is lazy initialized on first HoverEnter since window() only returns the real top-level widget after StatusTime has been inserted into the statusbar layout.

Let me know if you'd prefer a different parent setup.

tbertels:
> The tooltip is parented to window() instead of StatusTime. The reason is that Tooltip parents its internal QLabel to whatever you pass in, and child widgets get clipped to their parent's geometry. StatusTime is only status-bar sized, so the tooltip would render clipped. Using window() gives MainWindow as the parent (same Qt auto-cleanup, wide enough drawing surface).
>
> The tooltip is lazy initialized on first HoverEnter since window() only returns the real top-level widget after StatusTime has been inserted into the statusbar layout.

It looks fine to me!
Could you fixup the first commit and edit it to add the slightly reworded version of your explanation I've quoted?

Thoxy67:
Done! 

Squashed the two commits into one and folded the tooltip parenting explanation into the commit message (slightly reworded for the commit context). 

Force-pushed to the branch.

tbertels:
Thanks!

tbertels (commented):
I didn't catch this: we need to include the whole "Played: %1" and "Remaining: %1" as translatable strings, like it's done for instance with "Subtitles delay: %1 ms".
That way, the left-right direction can be changed by translators and even the spacing between Played/Remaining and ":" like in French.
@Thoxy67 Can you do it or do I go ahead and do it?

PR #946, opened 2026-05-12
settingswindow: add searchable settings input

### LINKED ISSUE #751: [Feature request] Search in options
Recently I've seen quite a few apps doing this: adding a search bar for options to make it easier to find the setting you are looking for.
Firefox:
Image

VLC:

Image

Thoxy67:
There's so many options in the settings that I couldn't always find what I was looking for, so I added a small search bar at the top of the Options window.
It just filters the left tree as you type, matches against category names, setting labels, combo box entries, and even keyboard shortcuts (both action names and the keys themselves, so typing "space" or "play" both work).
While I was at it I also gave the right pane title a small header bar so it's a bit more obvious where you are, and made it follow the theme.

Translated the new placeholder to French. Other languages will show the English string until someone translates it.

Resolves #751 

Sorry this is the last one for today but I think would make life easier for many people

[Enregistrement d'écran_20260512_193818.webm](https://github.com/user-attachments/assets/34af77f2-1c82-4be4-82ba-a65772939eb4)

### DIFF
--- a/src/settingswindow.cpp
+++ b/src/settingswindow.cpp
@@ -6,6 +6,7 @@
 #include <QColorDialog>
 #include <QStyleHints>
 #include <QToolTip>
+#include <QToolButton>
 #include "logger.h"
 #include "platform/unify.h"
 #include "settingswindow.h"
@@ -353,6 +354,13 @@ SettingsWindow::SettingsWindow(QWidget *parent) :
     setupColorPickers();
     setupUnimplementedWidgets();
     ui->pageStack->installEventFilter(this);
+
+    buildPageSearchIndex();
+
+    int headerHeight = std::max(ui->pageLabel->sizeHint().height(),
+                                ui->optionsSearchField->sizeHint().height());
+    ui->pageLabel->setFixedHeight(headerHeight);
+    ui->optionsSearchField->setFixedHeight(headerHeight);
 }
 
 SettingsWindow::~SettingsWindow()
@@ -386,6 +394,7 @@ void SettingsWindow::updateLanguage()
     ui->retranslateUi(this);
     ui->videoPresetApplied->clear();
     takeKeyMap(acceptedKeyMap);
+    buildPageSearchIndex();
 }
 
 void SettingsWindow::setupPageTree()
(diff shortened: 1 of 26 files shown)


src/settingswindow.cpp
```diff
@@ -6,6 +6,15 @@
...
 #include <QToolTip>
+#include <QLabel>
+#include <QCheckBox>
+#include <QRadioButton>
+#include <QGroupBox>
+#include <QPushButton>
+#include <QTabWidget>
+#include <QComboBox>
+#include <QListWidget>
+#include <QToolButton>
```
tbertels: These aren't all needed.

src/settingswindow.cpp
```diff
@@ -353,6 +362,15 @@ SettingsWindow::SettingsWindow(QWidget *parent) :
...
     ui->pageStack->installEventFilter(this);
+
+    buildPageSearchIndex();
+
+    int headerHeight = std::max(ui->pageLabel->sizeHint().height(),
+                                ui->optionsSearchField->sizeHint().height());
+    ui->pageLabel->setMinimumHeight(headerHeight);
+    ui->pageLabel->setMaximumHeight(headerHeight);
+    ui->optionsSearchField->setMinimumHeight(headerHeight);
+    ui->optionsSearchField->setMaximumHeight(headerHeight);
```
tbertels: You can use setFixedHeight.

src/settingswindow.cpp
```diff
@@ -1253,18 +1274,92 @@ QList<MpvOption> SettingsWindow::parseMpvOptions(const QString &optionsInline) c
...
+
+    for (int i = 0; i < tree->topLevelItemCount(); ++i) {
+        QTreeWidgetItem *top = tree->topLevelItem(i);
+        bool anyChildVisible = false;
+
+        for (int j = 0; j < top->childCount(); ++j) {
+            QTreeWidgetItem *child = top->child(j);
+            QModelIndex childIndex = tree->indexFromItem(child);
+            int pageIdx = pageIndexForTreeItem(childIndex);
+            bool match = child->text(0).toLower().contains(needle)
```
tbertels: Instead of using systematically `toLower()`, you can use `contains(needle, Qt::CaseInsensitive)`.

Thoxy67:
Good catches, thanks! Pushed an update addressing all three:

- Dropped the unnecessary includes (only `QToolButton` actually had to stay, the rest were transitive via `ui_settingswindow.h`)
- Switched to `setFixedHeight()`
- Replaced the `toLower()` calls with `contains(needle, Qt::CaseInsensitive)`

tbertels:
Thanks for this!

PR #950, opened 2026-05-14
mainwindow: arm AB loop when B is set at current playback position

### LINKED ISSUE #949: AB loop doesn't activate when B is set at current playback position
### Bug

When I set point B while the player is past where I want to loop back from, the loop never kicks in. I have to seek back somewhere inside the A to B range myself before it starts looping.

### Steps to reproduce

1. Open a video.
2. Enable `Play > Loop > Repeat` (Ctrl+Backspace).
3. Start playing.
4. Press "Set Loop Start" (A) at some point.
5. Let it play for a few seconds.
6. Press "Set Loop End" (B).

Expected: it starts looping right away.
Actual: it just keeps playing past B. I have to seek somewhere between A and B for the loop to actually start.

### Why I think it happens

mpc-qt hands the loop points straight to mpv via `ab-loop-a` / `ab-loop-b` (see `setLoopPoints` in `src/mpvwidget.cpp:589`). mpv only triggers the loop when the playhead crosses B from inside the range. Since "Set B" uses the current playback time, the playhead is sitting right at B when the point is set, so mpv never sees it inside the range and never arms the loop.

### Is this intended?

Or should mpc-qt nudge the playhead back to A when this happens so the loop activates right away?

### Suggested fix

In `src/mainwindow.cpp`, in `on_actionPlayLoopEnd_triggered()` (line 3425), seek back to A if the playhead is already at or past B when B gets set with the loop enabled:

```cpp
void MainWindow::on_actionPlayLoopEnd_triggered()
{
    positionSlider_->setLoopB(mpvObject_->playTime());
    if (ui->actionPlayLoopUse->isChecked()) {
        double a = positionSlider_->loopA();
        double b = positionSlider_->loopB();
        mpvObject_->setLoopPoints(a, b);
        if (a >= 0 && b > a && mpvObject_->playTime() >= b)
            mpvObject_->setPlayTime(a);
    }
}
```

A matching guard in `on_actionPlayLoopStart_triggered()` would cover the mirror case.

Happy to send a PR if this direction sounds right.

Thoxy67:
When you press "Set Loop End" while playing, B gets set to the current playback time, so the playhead is sitting right on B at the moment the loop is armed. mpv only triggers ab-loop when the playhead crosses B from inside the range, so playback just keeps going past B until you seek back into [A, B] yourself.

Fix is to seek back to A with setTime() right after setLoopPoints when the loop is enabled, the range is valid, and the playhead is at or past B. Applied to all three handlers that arm the loop (Set Start, Set End, and Use toggled), so the same situation is also handled when re-enabling the loop after playback drifted past B while it was off.

Resolves #949

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -3420,16 +3420,14 @@ void MainWindow::on_actionPlayLoopStart_triggered()
 {
     positionSlider_->setLoopA(mpvObject_->playTime());
     if (ui->actionPlayLoopUse->isChecked())
-        mpvObject_->setLoopPoints(positionSlider_->loopA(),
-                                  positionSlider_->loopB());
+        armAbLoop();
 }
 
 void MainWindow::on_actionPlayLoopEnd_triggered()
 {
     positionSlider_->setLoopB(mpvObject_->playTime());
     if (ui->actionPlayLoopUse->isChecked())
-        mpvObject_->setLoopPoints(positionSlider_->loopA(),
-                                  positionSlider_->loopB());
+        armAbLoop();
 }
 
 void MainWindow::on_actionPlayLoopUse_toggled(bool checked)
@@ -3441,11 +3439,22 @@ void MainWindow::on_actionPlayLoopUse_toggled(bool checked)
         if (positionSlider_->loopB() < 0 )
             positionSlider_->setLoopB(mpvObject_->playLength());
 
-        mpvObject_->setLoopPoints(positionSlider_->loopA(),
-                                  positionSlider_->loopB());
+        armAbLoop();
     }
     else
         mpvObject_->setLoopPoints(-1, -1);
+}
+
+void MainWindow::armAbLoop()
+{
+    double a = positionSlider_->loopA();
+    double b = positionSlider_->loopB();
+    mpvObject_->setLoopPoints(a, b);
+    if (a < 0 || b < 0)
+        return;
+    double loopEnd = std::max(a, b);
+    if (mpvObject_->playTime() > loopEnd)
+        mpvObject_->setTime(std::min(a, b));
 }
 
 void MainWindow::on_actionPlayLoopClear_triggered()
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -125,6 +125,7 @@ private:
     void raiseWindow();
     void createAlwaysOnTopWindow();
     void showAlwaysOnTopWindow(bool show);
+    void armAbLoop();
 
     QIcon createIconFromSvg(const QString &svgPath, int maxSize) const;
     QPixmap renderPixmapFromSvg(const QString &path) const;


src/mainwindow.cpp
```diff
@@ -3419,17 +3419,25 @@ void MainWindow::on_actionMoveSubtitlesDown_triggered()
...
-    if (ui->actionPlayLoopUse->isChecked())
-        mpvObject_->setLoopPoints(positionSlider_->loopA(),
-                                  positionSlider_->loopB());
+    if (ui->actionPlayLoopUse->isChecked()) {
+        double a = positionSlider_->loopA();
+        double b = positionSlider_->loopB();
+        mpvObject_->setLoopPoints(a, b);
+        if (a >= 0 && b > a && mpvObject_->playTime() >= b)
+            mpvObject_->setTime(a);
+    }
```
tbertels: It seems like a good candidate for a function since the same code is used three times.
And hopefully, the function name will self document the code.

mpv doesn't actually care about the order of loopA and loopB.

Thoxy67: Good call, pushed the refactor.

Extracted the duplicated setLoopPoints plus snap-back logic into a private armAbLoop() helper, so the three call sites (Set Start, Set End, Use toggled) just call it after updating their respective loop point. The function name carries the intent.

Also dropped the b > a assumption per your note. Used std::max(a, b) for the "past the loop end" check and std::min(a, b) for the seek target, so the snap-back behaves correctly regardless of which order A and B were set in.

src/mainwindow.cpp
```diff
@@ -3441,13 +3439,24 @@ void MainWindow::on_actionPlayLoopUse_toggled(bool checked)
...
 
+void MainWindow::armAbLoop()
+{
+    double a = positionSlider_->loopA();
+    double b = positionSlider_->loopB();
+    mpvObject_->setLoopPoints(a, b);
+    if (a < 0 || b < 0)
+        return;
+    double loopEnd = std::max(a, b);
+    if (mpvObject_->playTime() >= loopEnd)
```
tbertels: It's a nitpick, but as mentioned in the issue: if the player is paused, it does work if we set the loop /start,end and enable repeat (and the other way around). It's still considered in the bounds.
So `>= loopEnd` can be replaced by `> loopEnd` and the first commit comment can be adjusted.

tbertels (changes requested):

Thoxy67:
Changed >= loopEnd to > loopEnd in armAbLoop() so the snap back only fires when the playhead has actually drifted past B. Sitting exactly at B stays in bounds and lets mpv handle it on its own, paused or playing.

Also reworded the first commit message to drop the "playhead sitting right on B" framing and describe the real failure mode (playback drifting past B before mpv reacts to the freshly armed loop). Force-pushed since I rebased to reword.

tbertels (approved):
Can you fixup the first commit with the second one?
As mentioned, you don't need to push a new commit for requested changes, GitHub shows a "Compare" button next to the force-push line so it's easy to check changes.

Thoxy67:
Done, rebased on master and fixupped the two commits into one. The branch is now a single arm AB loop when B is set at current position commit on top of latest master, no merge commit in the way. The Compare button on the force-push line should show the squash + the >=  to >  change.

tbertels:
> The Compare button on the force-push line should show the squash + the >= to > change.

It doesn't actually, because it only shows actual changes between forces-pushes, and that change was already shown in [the one before](https://github.com/mpc-qt/mpc-qt/compare/b166d2542b9c0f8231f3b4fe50a137ca8ee1d9f4..ba384e10f6b40045fd37c72c24f739ee5940e6ac).

tbertels:
Thanks!

PR #985, opened 2026-06-01
Fix gapless playback-related issues, prevent mpv error when clearing filters

### LINKED ISSUE #982: Toggling yadif causes mpc-qt to overwrite the wrong playlist entry
This is the weirdest bug I have ever seen.  It's probably always been there, but recent changes have brought it to the fore.

To reproduce:
1. Load up a playlist with videos (I used [Trix's release of You And I Are Polar Opposites](https://nyaa.si/view/2097406))
2. Open the first video
3. Hold down the right arrow key until just before the end
4. Allow the video file to progress towards the end without input

Observation:
The third playlist entry has its metadata overwritten with the data from the second playlist entry.
The third playlist entry is bold indicating that the playlist window thinks the third entry is playing, even though the second entry is playing.
The file for the third playlist entry appears in the title bar.
Seeking to end of subsequent videos overwrites the n+1 entry with the data of the n entry, and shows the n+1 playlist entry in the title bar. The nth entry is the one actually playing.

For example, here's what the playlist looks like at the beginning of episode 4. (I've only watched this show up to the end of episode 3 before this bug got too annoying.)

Image

A git bisect led to bca3d3c1081d04b02160b788edc7b4b6b326d592, and more specifically the line
`emit videoFilter("yadif", "mode=1", enable);` (presently at line 999 of manager.cpp)
Commenting this line out restores proper playlist progression.
For what it's worth, hardware decoding was disabled.

So somewhere or other it appears that the state machine in manager.cpp is broken.

tbertels:
* mpvwidget: Only set filters when filter list isn't empty

> When combined with gapless playback (https://github.com/mpc-qt/mpc-qt/commit/a2d46911776d25ac22ce1a724e2d0487a1e35ac6), which uses the mpv playlist, trying to set empty filters causes mpv to skip one or more files.
> 
> Part 1 of fixing https://github.com/mpc-qt/mpc-qt/issues/982 ("Toggling yadif causes mpc-qt to overwrite the wrong playlist entry").
> 
> Fixes: https://github.com/mpc-qt/mpc-qt/commit/f8c35f2b678918e969c6b2b99f997fed0463e9c3 ("Add pre-defined audio and video filters support")
> Fixes: https://github.com/mpc-qt/mpc-qt/commit/a2d46911776d25ac22ce1a724e2d0487a1e35ac6 ("Implement gapless playback")

* Only clear filters when needed

> Avoid clearing filters when none are set, as mpv reports an error in that case.
>
> Fixes: 848f86dea7a24938cc9166486c1ec4022d5450fb ("mpvwidget: Clear audio/video filters before we set them")

* Only use gapless playback for audio-to-audio transitions

> Gapless playback relies on the mpv playlist, and is more prone to edge cases.
> 
> Besides the issue fixed by https://github.com/mpc-qt/mpc-qt/commit/0df29a39e53325371808800ccea21d2f78a13907, disabling video before loading a file and re-enabling it afterwards (https://github.com/mpc-qt/mpc-qt/commit/2868d2a1327abc02589e6e25b91e7d2362a5f4e4) causes mpv to report EOF for the new file.
> 
> Since gapless playback is only useful when transitioning between audio files, enable it only when opening an audio file while another audio file is already playing.
> 
> Part 2 of fixing https://github.com/mpc-qt/mpc-qt/issues/982 ("Toggling yadif causes mpc-qt to overwrite the wrong playlist entry").
> 
> Fixes: https://github.com/mpc-qt/mpc-qt/commit/2868d2a1327abc02589e6e25b91e7d2362a5f4e4 ("mpvwidget: Force mpv to clear last frame of previous video file")

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -726,7 +726,8 @@ void PlaybackManager::startPlayWithUuid(QUrl what, QUuid playlistUuid,
         emit startedPlayingFile(what);
     mpvObject_->fileOpen(what.isLocalFile() ? what.toLocalFile()
                                             : QUrl::fromPercentEncoding(what.toEncoded()),
-                         replaceMpvPlaylist);
+                         replaceMpvPlaylist,
+                         !videoList.isEmpty() && !videoListData[1].isImage);
     if (!with.isEmpty())
         mpvObject_->setSubFile(with.toString());
     mpvObject_->setPaused(playbackStartPaused);
--- a/src/mpvwidget.cpp
+++ b/src/mpvwidget.cpp
@@ -435,13 +435,15 @@ void MpvObject::setLogoBackground(const QColor &color)
 void MpvObject::setAudioFilters(const QList<QPair<QString, QString>> &filtersList)
 {
     emit ctrlCommand("no-osd af clear");
-    emit ctrlCommand("no-osd af set \"" + formatFiltersList(filtersList) + "\"");
+    if (!filtersList.empty())
+        emit ctrlCommand("no-osd af set \"" + formatFiltersList(filtersList) + "\"");
 }
 
 void MpvObject::setVideoFilters(const QList<QPair<QString, QString>> &filtersList)
 {
     emit ctrlCommand("no-osd vf clear");
-    emit ctrlCommand("no-osd vf set \"" + formatFiltersList(filtersList) + "\"");
+    if (!filtersList.empty())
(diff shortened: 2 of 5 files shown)


tbertels:
@cmdrkotori Can you confirm that it fixes https://github.com/mpc-qt/mpc-qt/issues/982 for you?
Once this is merged, I'll create the 26.07 branch.

cmdrkotori:
This fixes #982 for me, yes.

PR #1054, opened 2026-07-30
Add "Close / Minimize to tray" feature

### LINKED ISSUE #1053: Minimize to tray and keep running in tray?
Would it be possible to add a minimize to tray option?

Also would it be possible to keep running in the tray when the windows close button is pressed?

Thanks!

tbertels:
* Add "Close / Minimize to tray" feature

> Requires the tray icon and single window mode to be enabled.
> 
> The Exit command and middle-clicking the tray icon (newly added) always close MPC-QT.
> 
> Minimize to tray doesn't work (yet?) on Wayland in Wayland mode, it requires the Wayland compositor to let the window know that it's minimized.
> 
> Fixes https://github.com/mpc-qt/mpc-qt/issues/1053 ("Minimize to tray and keep running in tray?").

* mainwindow: Use same text for tray icon tooltip as the window title

> Especially useful to know what's playing when MPC-QT is hidden in the tray.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -731,6 +731,10 @@ void Flow::setupSettingsConnections()
     // settings -> mainwindow
     connect(settingsWindow, &SettingsWindow::trayIcon,
             mainWindow, &MainWindow::setTrayIcon);
+    connect(settingsWindow, &SettingsWindow::closeToTray,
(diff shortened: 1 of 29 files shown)


cmdrkotori:
This needs a rebase.

PR #1055, opened 2026-07-30
Update mpc-qt_ru.ts

Skif-off:
Some untranslated strings and some fixes.
(Sorry, I don't have a Weblate account.)

### DIFF
--- a/translations/mpc-qt_ru.ts
+++ b/translations/mpc-qt_ru.ts
@@ -5,7 +5,7 @@
     <name>About</name>
     <message>
         <source>Development Build</source>
-        <translation type="unfinished">Версия для разработчиков</translation>
+        <translation type="unfinished">Разрабатываемая версия</translation>
     </message>
     <message>
         <source>Version %1</source>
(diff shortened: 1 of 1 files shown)


tbertels:
The strings marked as `vanished` don't need to be unmarked as vanished or translated.
Qt lupdate (the `cmake --build build -t update_translations` command) marks them automatically as vanished when they're not used anymore in the code.
They're still kept in case they become useful again in the future, but Weblate doesn't show them to translators.

tbertels:
Thanks!
And no problem about not using Weblate, especially for major updates.
When there's only a few strings to update, it's easier to use Weblate (the account is free and you don't get spammed though you do receive notifications for new strings in projects you subscribe to).

Skif-off:
Thanks and I apologize for the inconvenience: I saw your message and started to correct file, then I was about to correct this PR, but you'd already done it.
I should have clarified the format specification and been more careful, but I just wanted to correct a few lines and I didn't notice how carried away I got... :)) (Looks like some part of the translation is machine translation, which can be bad or can be valid but not take context into account.)

P.S. I don't remember what problems I had with Weblate; it was a long time ago.

tbertels:
No need to apologize, it was just a matter of checking out your branch, running lupdate and force pushing the changes. 😉

tbertels:
> P.S. I don't remember what problems I had with Weblate; it was a long time ago.

Maybe it was the fact that Weblate didn't let you use a generated GitHub email.
That's not the case anymore:
image

PR #1072, opened 2026-08-15
Avoid showing the menu bar after the rest of the window on X11/XWayland startup

freeorca:
## Problem

On X11/XWayland, the first child paints and the window-manager decorations
can arrive asynchronously. The menu bar may appear slightly after the rest of
the window, which makes startup feel less crisp.

## Change

In `WindowManager::restoreAppWindow()`, when running on `xcb` and not
restoring a maximized or fullscreen window, map the window with opacity 0 and
restore its opacity 50 ms later.

Systems without compositing ignore the opacity value, so they are unaffected.

This is intentionally not enabled on native Wayland, where the delayed reveal
did not improve the issue reliably and would only delay startup.

The default window size fix has been split into #1079.

## Testing

- KDE Wayland / XWayland: 10/10 launches opened in one step, with no visible
  menu-bar delay.
- KDE Wayland / native Wayland: the delayed reveal is disabled there; 10/10
  launches also opened in one step.
- On current master, the test system still shows a very slight menu-bar
  delay, while this patch makes startup consistently crisp in the same
  environment.

### DIFF
--- a/src/platform/windowmanager.cpp
+++ b/src/platform/windowmanager.cpp
@@ -3,6 +3,7 @@
 #include <QMetaMethod>
 #include <QGuiApplication>
 #include <QStyle>
+#include <QTimer>
 #include <QWidget>
 #include "helpers.h"
 #include "logger.h"
@@ -98,6 +99,19 @@ void WindowManager::restoreAppWindow(MainWindow *window, const CliInfo &cliInfo)
     if (cliInfo.validCliPos)
         desiredPlace = cliInfo.cliPos;
 
+    // On X11/XWayland the first child paints and the window-manager
+    // decorations can arrive asynchronously, which may show the menu bar
+    // shortly after the rest of the window.  Map the window transparent
+    // and reveal it once the first complete frame is ready.  Systems without
+    // compositing ignore the opacity value.
+    QVariantMap savedState = data[keyState].toMap();
+    bool delayedReveal = QGuiApplication::platformName() == "xcb"
+            && !data.value(keyMaximized, false).toBool()
+            && !savedState.value("actionViewFullscreen", false).toBool();
+
+    if (delayedReveal)
+        window->setWindowOpacity(0.0);
+
     window->setGeometry(QRect(desiredPlace, desiredSize));
 
     if (data.value(keyMaximized, false).toBool())
@@ -106,6 +120,13 @@ void WindowManager::restoreAppWindow(MainWindow *window, const CliInfo &cliInfo)
         window->show();
     window->raise();
     window->setState(data[keyState].toMap());
+
+    if (delayedReveal) {
+        QTimer::singleShot(50, window, [window]() {
+            window->setWindowOpacity(1.0);
+            window->update();
+        });
+    }
 }
 
 void WindowManager::restoreDocks(QMainWindow *dockHost, QList<QDockWidget *> dockWidgets)


cmdrkotori:
Thanks for this. Instead of changing the opacity you can try using setAttribute(Qt::WA_DontShowOnScreen).  I believe at one stage we were going something like

```
setAttribute(Qt::WA_DontShowOnScreen);
show();
QTimer::singleShot(50, [this] {
    hide();
    setAttribute(Qt::WA_DontShowOnScreen, false);
    updateSize();
    show();
}
```

Although it has been quite some time (over half a decade) since we were touching this attribute, so don't take this code snippet as gospel.

Anyway, LGTM.

tbertels:
Thank for the PR!

Was an LLM involved? The code seems overly complicated for what it achieves.

If the only aim is for the window to have the noVideoSize_ on start, updating the default width and height in mainwindow.ui is enough.

tbertels:
@cmdrkotori About `Qt::WA_DontShowOnScreen`, I've tried using it without success to avoid showing the menubar and controls too small for a split second in Wayland mode with fractional scaling. This bug also explains why the logo is too small on start before the first window resize.
 
By the way, the code you're talking about is still there in some form:
https://github.com/mpc-qt/mpc-qt/blob/602966507b2ab9a3e89ca2d400112e6218a266d8/src/mainwindow.cpp#L69-L78

cmdrkotori:
With wayland mode on kde we get initial window centering for free, so it all works out of the box with this PR.  One thing I noticed that on Windows however is that this PR breaks preserving the size of the playlist window.  Not a problem on Linux though.

freeorca:
Thanks for the review!

Yes, an LLM assisted with drafting and iterating on this patch. The behavior
described under Testing was verified manually on KDE Wayland/XWayland.

Regarding `mainwindow.ui`: I considered that approach, but a fixed pixel size
only matches one configuration. The intended default is `noVideoSize_` plus the
real window chrome, which depends on the saved panel state and the
platform/theme. A fixed value would also not account for the width taken by
the playlist dock when the playlist is visible. On the test system the default
happens to be 500×406 with the playlist hidden, but that value is not
universal.

Regarding `Qt::WA_DontShowOnScreen`: the constructor already uses it to settle
the layout, but as you noted it does not prevent the first visible frame from
showing menus/controls at the intermediate size on Wayland with fractional
scaling. The short transparent-mapping period was the approach that gave
consistent first frames in our repeated testing, but I'm open to replacing it
if there is a better alternative.

I can't test Windows here. Would it be acceptable to restrict the new restore
path to `xcb` and `wayland` so Windows keeps its existing behavior, or would
you prefer the default-size fix to be cross-platform? I'm happy to update the
PR accordingly.

tbertels:
Some LLMs tend to add code during each iteration so we end up for a lot of unneeded code.
For instance, the `window->setWindowOpacity` calls aren't actually needed.
And I suspect we can remove more code while keeping the intended result.

Other than that, this approach (compared to just modifying the default size in mainwindow.ui) would make it possible to add an option to set the default window size.

tbertels:
> Make the panel visibility handlers use unconditional `setVisible()`, so
the saved state is applied correctly while the window is still hidden.

That part isn't needed either, but it does simplify the code while most likely having no meaningful performance impact, so I've moved it to a separate commit.

cmdrkotori:
>I can't test Windows here. Would it be acceptable to restrict the new restore
path to xcb and wayland so Windows keeps its existing behavior, or would
you prefer the default-size fix to be cross-platform? I'm happy to update the
PR accordingly.

It would probably be better to work out why this is approach is not working on Windows, rather than have two divergent code paths.

cmdrkotori:
>I can't test Windows here.

What I do is run MSYS2 under Windows 10 in a VirtualBox. With guest additions you get enough hardware acceleration to run mpc-qt.

freeorca:
Thanks for simplifying the branch. I retested the current PR head
(`0bfb8300`) on KDE Wayland/XWayland.

The size result is correct, but the menu bar appears after the rest of the
window in 9 out of 10 launches. This makes startup feel less crisp than the
current release, where the whole window appears at once; the late menu bar
gives the impression that the UI is still being assembled after the window
is already visible.

We saw the same behavior while developing this patch: the version without
the delayed reveal had the menu bar showing up after the rest of the window.
That is why the opacity-based delayed reveal was added.

To be transparent about its limits: 50 ms improved consistency a lot, but it
is not a guarantee that the menu bar's first paint is ready. The value was
chosen as a balance; we did not measure whether larger values would avoid
the issue or add a noticeable startup delay.

I also noticed one possible regression from the simplification: when the
saved state contains `actionViewFullscreen = true`, `setState()` is now
skipped before `show()` and is never called afterwards. Is that intentional?

tbertels:
I hadn't noticed that menubar bug on X11/XWayland.

The removal of actionViewFullscreen does prevent the fullscreen state from being remembered (when closing MPC-QT while in fullscreen).

And with these removal commits, this initial window size fix doesn't work anymore on Windows.

> One thing I noticed that on Windows however is that this PR breaks preserving the size of the playlist window. Not a problem on Linux though.

It only happens if we don't close the playlist window before closing MPC-QT.

tbertels:
I've reverted the removal commits.

As an alternative, here's a simpler version of the default window size fix: https://github.com/tbertels/mpc-qt/commit/42c7701a5a8e2c0a842e0fd57ce7552ad173971e

It uses the size defined in mainwindow.ui for first start size, and saves the UI window chrome size on each app stop for the following start.

tbertels:
> As an alternative, here's a simpler version of the default window size fix: [tbertels@42c7701](https://github.com/tbertels/mpc-qt/commit/42c7701a5a8e2c0a842e0fd57ce7552ad173971e)
> 
> It uses the size defined in mainwindow.ui for first start size, and saves the UI window chrome size on each app stop for the following start.

I've created #1079 for that.

This PR can then focus on preventing the menu bar from showing up after the rest of the window.
The approach of this PR only works on X11/XWayland, so it should only run on it to avoid slowing down the start in Wayland mode.

tbertels:
> https://github.com/mpc-qt/mpc-qt/blob/602966507b2ab9a3e89ca2d400112e6218a266d8/src/mainwindow.cpp#L69-L78

@cmdrkotori  Do you remember which issue was fixed by this code (added by 167072ed2138200d769fc3cb57b567ce791595e7)?

I've tried earlier to remove it and there was an obvious regression on both X11/XWayland and Wayland.
That doesn't seem to be the case anymore. So it looks like there was a bug fix in either Qt or KDE.

tbertels (approved):
Nice, much neater!

cmdrkotori:
> > https://github.com/mpc-qt/mpc-qt/blob/602966507b2ab9a3e89ca2d400112e6218a266d8/src/mainwindow.cpp#L69-L78
> 
> @cmdrkotori Do you remember which issue was fixed by this code (added by [167072e](https://github.com/mpc-qt/mpc-qt/commit/167072ed2138200d769fc3cb57b567ce791595e7))?

It was exposing the window to the windowing system so things like window chrome could be reliably calculated.

tbertels:
> > > https://github.com/mpc-qt/mpc-qt/blob/602966507b2ab9a3e89ca2d400112e6218a266d8/src/mainwindow.cpp#L69-L78
> > 
> > 
> > @cmdrkotori Do you remember which issue was fixed by this code (added by [167072e](https://github.com/mpc-qt/mpc-qt/commit/167072ed2138200d769fc3cb57b567ce791595e7))?
> 
> It was exposing the window to the windowing system so things like window chrome could be reliably calculated.

I've checked for regressions with a Qt 6.5 Flatpak, on KDE Wayland, X11, and Cinnamon but didn't notice anything with window auto-resize on and off.
Same thing with Kubuntu 24.04 (Plasma 5.27) and Ubuntu 22.04.
I've also checked with older MPC-QT versions (down to 24.12), nothing either.
So it may have been a bug in Qt < 6.5.

PR #1076, opened 2026-08-20
playlistwindow: Append file extension when exporting playlist

polipoki-ki:
On GNOME and GTK-based DEs, the extension is not automatically added when exporting a playlist.
Because of this, when exporting/importing a playlist, playlists with the correct MIME type but without an extension are not displayed in the file dialog.

### DIFF
--- a/src/playlistwindow.cpp
+++ b/src/playlistwindow.cpp
@@ -6,6 +6,8 @@
 #include <QInputDialog>
 #include <QFileDialog>
 #include <QMenu>
+#include <QFileInfo>
+#include <QMessageBox>
 #include "logger.h"
 #include "playlistwindow.h"
 #include "ui_playlistwindow.h"
@@ -763,14 +765,42 @@ void PlaylistWindow::finishSearch()
 void PlaylistWindow::savePlaylist(const QUuid &playlistUuid)
 {
     static QFileDialog::Options options = QFileDialog::Options();
+
 #ifdef Q_OS_MAC
     options = QFileDialog::DontUseNativeDialog;
 #endif
-    QString file;
-    file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), QString(),
-                                        tr("Playlist files (*.m3u *.m3u8)"), nullptr, options);
+
+    QString file = tr("Playlist") + ".m3u";
+    bool confirmOverwrite = false;
+
+    while (true) {
+        file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), file,
+                                            tr("Playlist files") + " (*.m3u *.m3u8)", nullptr, options);
+        if(file.isEmpty())
+            return;
+
+        if (!file.endsWith(".m3u")) {
+            file.append(".m3u");
+            confirmOverwrite = true;
+        }
+
+        QFileInfo info(file);
+        if (confirmOverwrite && info.exists()) {
+            QMessageBox msgBox(this);
+            msgBox.setIcon(QMessageBox::Warning);
+            msgBox.setWindowTitle(tr("Export Playlist File"));
+            msgBox.setText(tr("File \"%1\" already exists.\nDo you want to replace it?").arg(info.fileName()));
+            msgBox.setStandardButtons(QMessageBox::Ok | QMessageBox::Cancel);
+
+            if (msgBox.exec() != QMessageBox::Ok) {
+                confirmOverwrite = false;
+                continue;
+            }
+        }
+        break;
+    }
     auto pl = PlaylistCollection::getSingleton()->getPlaylist(playlistUuid);
-    if (!file.isEmpty() && pl)
+    if (pl)
         emit exportPlaylist(file, pl->toStringList());
 }
 
--- a/translations/mpc-qt_ar.ts
+++ b/translations/mpc-qt_ar.ts
@@ -2042,10 +2042,6 @@ No action will be triggered.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation type="unfinished"></translation>
-    </message>
-    <message>
         <source>Open</source>
         <translation type="unfinished"></translation>
     </message>
@@ -2087,6 +2083,15 @@ No action will be triggered.</source>
     </message>
     <message>
         <source>Enter playlist name</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Playlist files</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>File &quot;%1&quot; already exists.
+Do you want to replace it?</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
--- a/translations/mpc-qt_ca.ts
+++ b/translations/mpc-qt_ca.ts
@@ -2261,7 +2261,7 @@ No s&apos;activarà cap acció.</translation>
     </message>
     <message>
         <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation>Fitxers de llista de reproducció (*.m3u *.m3u8)</translation>
+        <translation type="vanished">Fitxers de llista de reproducció (*.m3u *.m3u8)</translation>
     </message>
     <message>
         <source>Export File</source>
@@ -2318,6 +2318,15 @@ No s&apos;activarà cap acció.</translation>
     <message>
         <source>Enter playlist name</source>
         <translation>Introduir nom de la llista de reproducció</translation>
+    </message>
+    <message>
+        <source>Playlist files</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>File &quot;%1&quot; already exists.
+Do you want to replace it?</source>
+        <translation type="unfinished"></translation>
     </message>
     <message>
         <source>&amp;New Playlist</source>
--- a/translations/mpc-qt_cs.ts
+++ b/translations/mpc-qt_cs.ts
@@ -2281,7 +2281,7 @@ No action will be triggered.</source>
     </message>
     <message>
         <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation>Playlistové soubory (*.m3u *.m3u8)</translation>
+        <translation type="vanished">Playlistové soubory (*.m3u *.m3u8)</translation>
     </message>
     <message>
         <source>Export File</source>
@@ -2338,6 +2338,15 @@ No action will be triggered.</source>
     <message>
         <source>Enter playlist name</source>
         <translation>Zadejte název playlistu</translation>
+    </message>
+    <message>
+        <source>Playlist files</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>File &quot;%1&quot; already exists.
+Do you want to replace it?</source>
+        <translation type="unfinished"></translation>
     </message>
     <message>
         <source>&amp;New Playlist</source>
--- a/translations/mpc-qt_de.ts
+++ b/translations/mpc-qt_de.ts
@@ -2297,7 +2297,7 @@ Es wird keine Aktion ausgelöst.</translation>
     </message>
     <message>
         <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation>Playlist-Dateien (*.m3u *.m3u8)</translation>
+        <translation type="vanished">Playlist-Dateien (*.m3u *.m3u8)</translation>
     </message>
     <message>
         <source>Open</source>
@@ -2350,6 +2350,15 @@ Es wird keine Aktion ausgelöst.</translation>
     <message>
         <source>Enter playlist name</source>
         <translation>Playlist-Name eingeben</translation>
+    </message>
+    <message>
+        <source>Playlist files</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>File &quot;%1&quot; already exists.
+Do you want to replace it?</source>
+        <translation type="unfinished"></translation>
     </message>
     <message>
         <source>&amp;New Playlist</source>
--- a/translations/mpc-qt_en.ts
+++ b/translations/mpc-qt_en.ts
@@ -2299,7 +2299,7 @@ No action will be triggered.</source>
     </message>
     <message>
         <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation>Playlist files (*.m3u *.m3u8)</translation>
+        <translation type="vanished">Playlist files (*.m3u *.m3u8)</translation>
     </message>
     <message>
         <source>Export File</source>
@@ -2356,6 +2356,15 @@ No action will be triggered.</source>
     <message>
         <source>Enter playlist name</source>
         <translation>Enter playlist name</translation>
+    </message>
+    <message>
+        <source>Playlist files</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>File &quot;%1&quot; already exists.
+Do you want to replace it?</source>
+        <translation type="unfinished"></translation>
     </message>
     <message>
         <source>&amp;New Playlist</source>
--- a/translations/mpc-qt_es.ts
+++ b/translations/mpc-qt_es.ts
@@ -2143,7 +2143,7 @@ No action will be triggered.</source>
     </message>
     <message>
         <source>Playlist files (*.m3u *.m3u8)</source>
-        <translation>Archivos de lista de reproducción (*.m3u *.m3u8)</translation>
+        <translation type="vanished">Archivos de lista de reproducción (*.m3u *.m3u8)</translation>
     </message>
     <message>
         <source>Export File</source>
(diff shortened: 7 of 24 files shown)


tbertels:
Thanks for the PR!

It should already work right now.

Here's an example when entering "test" in the file save dialog.
The file is already saved as test.m3u and if I try to save it again under that name, there's an error (already translated as it's a native dialog).

Here's on KDE Plasma:

Screenshot_20260820_153737

And on Windows 10:

VirtualBox_Windows 10_20_08_2026_15_31_40

polipoki-ki:
> t should already work right now.
> 
> Here's an example when entering "test" in the file save dialog. The file is already saved as test.m3u and if I try to save it again under that name, there's an error (already translated as it's a native dialog).
> 
> Here's on KDE Plasma:

If the "Automatically select filename extension (.m3u)" checkbox is unchecked, the file is saved without an extension. 

In GNOME, a playlist is also created without an extension.

(Export PLaylist in Gnome)
Снимок экрана от 2026-08-20 18-00-15

(Import Playlist in Gnome)
Снимок экрана от 2026-08-20 18-01-11

 Consequently, during export or import, these playlists are not displayed in the file dialog, despite having the correct MIME type.

tbertels:
It looks like [QFileDialog::setDefaultSuffix](https://doc.qt.io/qt-6.8/qfiledialog.html#defaultSuffix-prop) could be used for that:

> Suffix added to the filename if no other suffix was specified.
> 
> This property specifies a string that is added to the filename if it has no suffix yet. The suffix is typically used to indicate the file type (e.g. "txt" indicates a text file).
> 
> If the first character is a dot ('.'), it is removed.

polipoki-ki:
> It looks like [QFileDialog::setDefaultSuffix](https://doc.qt.io/qt-6.8/qfiledialog.html#defaultSuffix-prop) could be used for that:
> 
> > Suffix added to the filename if no other suffix was specified.
> > This property specifies a string that is added to the filename if it has no suffix yet. The suffix is typically used to indicate the file type (e.g. "txt" indicates a text file).
> > If the first character is a dot ('.'), it is removed.

Yes, I am aware of QFileDialog::setDefaultSuffix. 

I chose not to use it for the following reasons: 

1. setDefaultSuffix only supports a single file extension.
If you need to support multiple extensions, this approach becomes problematic. 

In my implementation, you can add additional supported extensions  in `namefilters` if needed.

  ```
QString file, selectedFilter;
    QStringList filters, namefilters;

    namefilters << "*.m3u" << "*.m3u8";
    filters << tr("Playlist files") + QStringLiteral(" (%1)").arg(namefilters.join(QChar::Space));
    selectedFilter = filters.at(0);
```
    
2. setDefaultSuffix only checks for the presence of an extension, but does not compare it with the list of available (allowed) extensions.

**Comment by tbertels** [comment]
This approach is less user friendly than how it currently works on both KDE (and likely on other Qt-based DEs) and Windows.

That's because it disables the native file overwrite confirmation - shown while still inside the file dialog - with [QFileDialog::DontConfirmOverwrite](https://doc.qt.io/qt-6.8/qfiledialog.html#Option-enum) seems like a regression outside of GTK-based DEs.

Note that [GTK developers don't want to implement auto file extension](https://gitlab.gnome.org/GNOME/gtk/-/work_items/5367), and instead want app developers to create their own custom (and thus non native) file dialogs for file export.

So, I suggest:
- ~Checking if QFileDialog::setDefaultSuffix actually works (for the m3u extension, which is fine here right now) and does show a native overwrite dialog. Though even if the native overwrite dialog is shown, that solution doesn't check for a valid extension.~
- ~If the above doesn't work,~ check if the returned filename contains a valid extension, and if not:
- - Add the "m3u" extension
- - Check if the file already exists and show a custom overwrite dialog if needed

**Comment by cmdrkotori** [comment]
>Note that [GTK developers don't want to implement auto file extension](https://gitlab.gnome.org/GNOME/gtk/-/work_items/5367), and instead want app developers to create their own custom (and thus non native) file dialogs for file export.

Sigh.  That's not hard at all.  At least in C++ with any sufficiently advanced toolkit.  In any case we should be wary of making fixes to cover something that may eventually unbreak in the future.  Specially so if this fix could become unnecessary at any arbitrary point.  (whenever gtk devs get off their ass and try to tackle something remotely uncomplicated.)  I also don't actively test how GNOME integrates with mpc-qt.  I usually just assume the environment will try to ape the conventions established long ago and do something sane, like Qt seems to do.

That said, I'm not against making sure that the file extension was returned, and if not, adding it with a possible check, which seems to be what this PR is aiming at.  I'm not so sure that disabling confirm file overwrite for everyone is the right approach though.  Perhaps a conditional check for XDG_SESSION_DESKTOP=GNOME would be appropriate, and if so opt-in to the new behavior.

**Comment by tbertels** [comment]
I've tried with `QFileDialog::setDefaultSuffix`, it's not useful: even on KDE, it only adds the extension when returning the filename, so no overwrite prompt is shown. And as mentioned, it doesn't check for an invalid extension.

On KDE, users can still uncheck the checkbox to auto-add the extension, so on both GTK-based DEs and on KDE, it's useful to check for a valid extension.

We could use a while loop to show the dialog again, and making sure to restore the right `dir`.
It's also useful to use `tr("Playlist") + ".m3u"` instead of `QString()` for the default dir (on KDE the name without the extension is selected, so the user can directly type a different the name). It also informs the user of the correct extension.

So we could:
- Use a while loop
- Pre-set `file` to  `tr("Playlist") + ".m3u"` and use it in the dialog for the `dir` parameter
- Check that there's a valid extension (`.m3u`)
- if not:
- - Add the `.m3u` extension to `file`
- - Check that the file doesn't exist already
- - if it does:
- - - Ask the user if it should be overwritten
- - - If the user refuses:
- - - - Show the dialog again, with the corrected `file` as `dir` parameter

File: src/playlistwindow.cpp [hunk 1]
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
 void PlaylistWindow::savePlaylist(const QUuid &playlistUuid)
 {
     static QFileDialog::Options options = QFileDialog::Options();
+    options = QFileDialog::DontConfirmOverwrite;
+
```
tbertels, line 768: QFileDialog::DontConfirmOverwrite shouldn't be used as it degrades the user experience where it's supported.
A `confirmOverwrite` variable needs to be used to only show the overwrite prompt when we have had to add the extension.

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
-    file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), QString(),
-                                        tr("Playlist files (*.m3u *.m3u8)"), nullptr, options);
-    auto pl = PlaylistCollection::getSingleton()->getPlaylist(playlistUuid);
-    if (!file.isEmpty() && pl)
-        emit exportPlaylist(file, pl->toStringList());
+
+    QString file, filter, dir;
+
+    filter = tr("Playlist files") + " (*.m3u)";
+    dir = QStandardPaths::writableLocation(QStandardPaths::HomeLocation);
```
tbertels: file, filter, dir should be declared when first assigned.

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
+
+    while (true)
+    {
+        file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), dir + "/" + tr("Playlist") + ".m3u", filter, &filter, options);
```
tbertels: The full dir + filename (with added .m3u extension on second loop) - which is simply the variable `file` - should be reused so that the filename entered by the user isn't lost when it already exists.

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
+        auto pl = PlaylistCollection::getSingleton()->getPlaylist(playlistUuid);
```
tbertels: This should be just above the emit.

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
+            QFileInfo info(file);
+            // File extensions have been converted to lowercase for display in the file dialog during import and export
+            if (info.suffix().compare("m3u", Qt::CaseSensitive) != 0)
+            {
+                file = info.absolutePath() + "/" + info.completeBaseName() + "." + info.suffix().toLower();
+                info.setFile(file);
+            }
+
+            // Returns true if the file system entry this QFileInfo refers to exists; otherwise returns false
+            if(info.exists())
```
tbertels: This needs a second condition like `confirmOverwrite` which is true when the file picker didn't provide a valid extension (meaning that the user hasn't been asked already or at least not for the actual file with correct extension).

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
+
+        if (!file.isEmpty() && pl)
+        {
+            QFileInfo info(file);
+            // File extensions have been converted to lowercase for display in the file dialog during import and export
+            if (info.suffix().compare("m3u", Qt::CaseSensitive) != 0)
+            {
+                file = info.absolutePath() + "/" + info.completeBaseName() + "." + info.suffix().toLower();
+                info.setFile(file);
+            }
```
tbertels: This could be replaced by 
```
        if (!file.endsWith(".m3u", Qt::CaseInsensitive)) {
            file.append(".m3u");
            confirmOverwrite = true;
        }
```

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
-    if (!file.isEmpty() && pl)
-        emit exportPlaylist(file, pl->toStringList());
+
+    QString file, filter, dir;
+
+    filter = tr("Playlist files") + " (*.m3u)";
+    dir = QStandardPaths::writableLocation(QStandardPaths::HomeLocation);
+
+    while (true)
+    {
```
tbertels: The `{`  should be on the same line as the condition.

tbertels (changes requested):
Thanks for the update!

src/playlistwindow.cpp
```diff
@@ -763,15 +767,66 @@ void PlaylistWindow::finishSearch()
...
-    QString file;
-    file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), QString(),
-                                        tr("Playlist files (*.m3u *.m3u8)"), nullptr, options);
-    auto pl = PlaylistCollection::getSingleton()->getPlaylist(playlistUuid);
-    if (!file.isEmpty() && pl)
-        emit exportPlaylist(file, pl->toStringList());
+
+    QString file, filter, dir;
+
+    filter = tr("Playlist files") + " (*.m3u)";
```
tbertels: There's no need to declare `filter` or use it with `&filter`, just keep `QFileDialog::getSaveFileName` as it was but replace `QString()` by `file` initialized with `tr("Playlist") + ".m3u"`.

tbertels (changes requested):

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+
+    while (true) {
+        file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), dir + "/" + namefile + ".m3u",
+                                            tr("Playlist files") + " (*.m3u *.m3u8)", nullptr, options);
+
+        if(file.isEmpty())
+            return;
+
+        if (!file.endsWith(".m3u", Qt::CaseInsensitive))
+        {
```
tbertels: Here's how the braces should be inside functions (not just for this line of code):
```
 if (condition) {
 } else {
 }
```

src/playlistwindow.cpp
polipoki-ki: If you don't check the extension's case and convert it to lowercase when necessary:
1. The file existence check will be incorrect.
This is because Linux treats two files with the same name but different cases (including extensions ) as separate files.
2. The file will be saved with the extension in the wrong case.
3. Files with incorrect case extensions will not appear in the file import and export dialogs in the program.

tbertels: 1. It won't strictly speaking be incorrect, because the other file won't be overwritten.
2. True
3. Indeed, on GNOME (KDE shows both), only the .m3u file is shown

So the simple fix is to use 
```
        if (!file.endsWith(".m3u")) {
            file.append(".m3u");
            confirmOverwrite = true;
        }
```
The user entering ".M3U" is an edge case, so adding ".m3u" to it should be fine (and the next time, the user won't make that mistake).
That's also how KDE handles it when the auto extension checkbox is checked.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
 void PlaylistWindow::savePlaylist(const QUuid &playlistUuid)
 {
     static QFileDialog::Options options = QFileDialog::Options();
+
 #ifdef Q_OS_MAC
     options = QFileDialog::DontUseNativeDialog;
 #endif
+
     QString file;
```
tbertels: `file` should be initialized with `tr("Playlist") + ".m3u"`.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
-    file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), QString(),
-                                        tr("Playlist files (*.m3u *.m3u8)"), nullptr, options);
+    QString dir = QStandardPaths::writableLocation(QStandardPaths::HomeLocation);
```
tbertels: `dir` isn't needed.
QStandardPaths::HomeLocation is the default path already.
And when we'll add an option in Export to set the default folder, that folder can simply be prepended to `file` during its initialization.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+    QString namefile = tr("Playlist");
+    bool confirmOverwrite = false;
+
+    while (true) {
+        file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), dir + "/" + namefile + ".m3u",
```
tbertels: `dir` and `namefile` aren't needed.
So `dir + "/" + namefile + ".m3u"` should be replaced by `file`.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+        if(file.isEmpty())
+            return;
+
+        if (!file.endsWith(".m3u", Qt::CaseInsensitive))
+        {
+            file.append(".m3u");
+            confirmOverwrite = true;
+        }
+
+        if (!file.isEmpty())
```
tbertels: This has already been checked above.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+        {
+            QFileInfo info(file);
+            //File extensions have been converted to lowercase for display in the file dialog during import and export
+            if (info.suffix().compare("m3u", Qt::CaseSensitive) != 0)
+            {
+                file = info.absolutePath() + "/" + info.completeBaseName() + "." + info.suffix().toLower();
+                info.setFile(file);
+                confirmOverwrite = true;
+            }
```
tbertels: As discussed, this isn't really needed.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+    bool confirmOverwrite = false;
+
+    while (true) {
+        file = QFileDialog::getSaveFileName(this, tr("Export Playlist File"), dir + "/" + namefile + ".m3u",
+                                            tr("Playlist files") + " (*.m3u *.m3u8)", nullptr, options);
+
+        if(file.isEmpty())
+            return;
+
+        if (!file.endsWith(".m3u", Qt::CaseInsensitive))
```
tbertels: As discussed, Qt::CaseInsensitive can be removed.

src/playlistwindow.cpp
```diff
@@ -763,12 +766,64 @@ void PlaylistWindow::finishSearch()
...
+            {
+                file = info.absolutePath() + "/" + info.completeBaseName() + "." + info.suffix().toLower();
+                info.setFile(file);
+                confirmOverwrite = true;
+            }
+
+            if (confirmOverwrite)
+            {
+                // Returns true if the file system entry this QFileInfo refers to exists; otherwise returns false
+                if(info.exists())
```
tbertels: These two conditions should be combined.

tbertels (changes requested):

tbertels:
About the third commit, "russhian" should be spelled "Russian".

tbertels:
I've removed the third `file.isEmpty()` check and two empty lines.
I've also changed back the language of the Russian .ts file to "ru" (from ru_RU) as Russian-speaking users of other countries can still use the Russian locale if no specific one exists.
I've also merged the strings update commit with the main commit and reworded it.

tbertels (approved):
Thanks again for the PR, it will be a very good improvement for GNOME and Cinnamon users!
Removing `(*.m3u *.m3u8)` from the translation is also a good idea.

polipoki-ki:
> Thanks again for the PR, it will be a very good improvement for GNOME and Cinnamon users! Removing `(*.m3u *.m3u8)` from the translation is also a good idea.

Thanks you too for your patience and detailed explanations.

PR #1096, opened 2026-09-02
settingswindow: Download auto-generated subtitles matching user options

tbertels:
YouTube returns an error when yt-dlp tries to download all (300+) auto-translated subtitles.
But downloading just the ones from Options > Playback > Default track preference > Subtitles is fine.

Also download auto-generated subtitles in the video language.

The way they are displayed isn't as good as normal subtitles (only one line is refreshed at a time).


tbertels:
See https://github.com/yt-dlp/yt-dlp/issues/1734 about the display issue.

PR #1102, opened 2026-09-04
videopreview: Add a drop shadow and fix 1-pixel black line on right side with fractional scaling

### LINKED ISSUE #482: Various video preview feature improvements
Dear friends!
Thank you very much for your continued improvement of MPC.
Please continue to improve the video preview feature.
I tried this excellent player very carefully tonight and found that its stability is very good! 👍 So I want to participate in your improvement plan, so I would like to make the following improvement suggestions:
1. The video preview screen image is not synchronized with the left and right mouse slide and hover;

2. Please allow users to customize the video preview screen image of any size;
Please allow users to customize the video preview screen image resolution (for example, 4K, 2K, 1080P, 720P, etc.). If the user's network speed is fast or not slow, then previewing the highest quality image will not be too slow (with the rapid popularization of 5G networks and ultra-large screen devices, we should look further ahead and consider how to develop and build the next generation of media playback platforms from now on... For example, the online video preview function in MPC-BE is a very successful case.). Therefore, we should give users more autonomy to choose how to use the video preview image resolution (for example, 4K, 2K, 1080P, 720P, etc.), rather than restricting and limiting it at the software level.
Image<br/>
Image
3. Please remove the black background of the video preview time pointed by the red arrow;
4. Add rounded corners and shadow effects (border glow effect) to the video preview screen image;
Image
Thank you again for providing an unparalleled media playback tool, the world will be a better place because of you! 🎉👏🏆👍😜

tbertels:
* videopreview: Add a drop shadow

> This is mostly cosmetic, but it helps differentiate the preview from the actual video.

* videopreview: Fix 1-pixel black line on right side with fractional scaling

> With some combinations of fractional scaling, aspect ratio and preview size setting, a 1-pixel black line was shown on the right side of the video preview.

### DIFF
--- a/src/widgets/videopreview.cpp
+++ b/src/widgets/videopreview.cpp
@@ -1,16 +1,25 @@
 #include <QApplication>
+#include <QGraphicsDropShadowEffect>
 #include "videopreview.h"
 
 static constexpr char logModule[] =  "videopreview";
 static constexpr int previewMarginX = 20;
 static constexpr int labelHeight = 20;
+static constexpr int shadowMargin = 8;
 
 VideoPreview::VideoPreview(QWidget *parent) : QWidget(parent)
 {
     mpv = new MpvObject(this);
-    videoWidget = new MpvGlWidget(mpv, parent);
+    videoContainer = new QWidget(parent);
+    auto *shadow = new QGraphicsDropShadowEffect(videoContainer);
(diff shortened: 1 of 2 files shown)


tbertels:
Here's before:

Capture d&#39;écran_20260904_160347

And after:

Capture d&#39;écran_20260904_160414

PR #1105, opened 2026-09-08
Add Hold-to-Speed Feature for Quick Playback Control

jimzrt:
This PR adds a new "Pressed Speed" feature that allows users to temporarily change the playback speed by holding down the left mouse button in the video window. When the button is released, playback returns to normal speed.

I personally was missing this feature - don't know if there is interest. I use it all the time.

### DIFF
--- a/src/settingswindow.ui
+++ b/src/settingswindow.ui
@@ -1213,6 +1213,38 @@ media file played</string>
                 </property>
                </widget>
               </item>
+              <item row="4" column="0">
+               <widget class="QLabel" name="playbackHoldSpeedLabel">
+                <property name="text">
+                 <string>Hold-to-speed</string>
+                </property>
+               </widget>
+              </item>
+              <item row="4" column="1">
+               <widget class="QDoubleSpinBox" name="playbackHoldSpeed">
+                <property name="sizePolicy">
+                 <sizepolicy hsizetype="MinimumExpanding" vsizetype="Fixed">
+                  <horstretch>0</horstretch>
+                  <verstretch>0</verstretch>
+                 </sizepolicy>
+                </property>
+                <property name="suffix">
+                 <string notr="true">x</string>
+                </property>
+                <property name="minimum">
+                 <double>1.000000000000000</double>
+                </property>
+                <property name="maximum">
+                 <double>100.000000000000000</double>
+                </property>
+                <property name="singleStep">
+                 <double>0.100000000000000</double>
+                </property>
+                <property name="value">
+                 <double>2.500000000000000</double>
+                </property>
+               </widget>
+              </item>
              </layout>
             </widget>
            </item>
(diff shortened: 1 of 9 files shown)


cmdrkotori:
The timeout feels too short and sometimes it breaks pausing when I repeatedly press my mouse.  Perhaps a longer timeout would work.

src/settingswindow.ui
```diff
@@ -1213,6 +1213,38 @@ media file played</string>
...
+              <item row="4" column="1">
+               <widget class="QDoubleSpinBox" name="playbackHoldSpeed">
+                <property name="sizePolicy">
+                 <sizepolicy hsizetype="MinimumExpanding" vsizetype="Fixed">
+                  <horstretch>0</horstretch>
+                  <verstretch>0</verstretch>
+                 </sizepolicy>
+                </property>
+                <property name="suffix">
+                 <string notr="true">x</string>
```
tbertels, line 1232: The "x" should be made translatable.

tbertels (changes requested):

tbertels:
> The timeout feels too short and sometimes it breaks pausing when I repeatedly press my mouse. Perhaps a longer timeout would work.

On YouTube, the timeout seems to be 500 ms (instead of 150 ms here right now).

A checkbox on the left to enable it (checked by default) would also be a good idea as [not everyone loves it](https://www.reddit.com/r/youtube/comments/17yocuj/that_stupid_click_and_hold_to_play_at_2x_speed/).

And finally, a key shortcut to use while not using the mouse, though that could be added afterwards.

jimzrt:
Thanks for the review. I amended the commit and force-pushed, so the previous review comments are on the old history. The current commit includes:

- The hold timeout is now 500 ms.
- Hold-to-speed is now a checkbox . I left it off by default. The speed box is disabled until the option is turned on.
- The x suffix is translatable.
- If the pointer moves far enough to drag the window before the timeout ends, the hold is cancelled.

A keyboard shortcut can be added later, as suggested.

src/settingswindow.ui
```diff
@@ -1213,6 +1213,44 @@ media file played</string>
                 </property>
                </widget>
               </item>
+              <item row="4" column="0">
+               <widget class="QCheckBox" name="playbackHoldSpeedEnable">
```
tbertels: The checkbox should have a
```
                <property name="sizePolicy">
                 <sizepolicy hsizetype="Minimum" vsizetype="Preferred">
                  <horstretch>0</horstretch>
                  <verstretch>0</verstretch>
                 </sizepolicy>
                </property>
```
 (like `playbackMouseHideFullscreen` for instance) so that it's vertically centered with the field on the right.
If you use Qt Widgets Designer, just set the vertical size policy to "Preferred".

tbertels (changes requested):
Could you add a commit description explaining simply what that feature does and how to use it?
About the commit title, you could add "feature" at the the end of "Add hold-to-speed".

Could you generate the translations and add them to the commit?
`cmake --build build -t update_translations`

Other than that and the minor change request, it looks fine to me.

PR #1115, opened 2026-09-14
mpris: Fix stale length after switching files

noctuum:
`maybeChangeMetadata()` detected changes with a bitfield of four presence flags, so it cannot see one value replaced by another. Across a file switch no flag moves, `metadata_` is never rebuilt, and `mpris:length` keeps the previous file's duration while `xesam:url` advances to the new one. A duration going from 0 to its real value is hidden the same way.

`buildMetadata()` now returns a fresh map and `maybeChangeMetadata()` compares it with the published one. `mpris:length` is omitted while the duration is unknown instead of published as 0.

Fixes: https://github.com/mpc-qt/mpc-qt/commit/789fdb87a35b19d73276ae1d781655f8e2c496f0 ("ipcmpris: message optimization")

Tested with Qt 6.12.0beta4 and mpv 0.41.0 on Arch Linux:

| Check | Result |
|---|---|
| Playlist switch from a 47:32 file to a 23:04 file | `mpris:length` reports 23:04 |
| Stop, then play the same file again | key absent while loading, then 23:04 |
| Seek inside a file | length unchanged |
| Jump to a non-adjacent playlist entry | 28:23, matches `ffprobe` |
| 42 s of steady playback | 32 `PropertiesChanged`, none carrying `Metadata` |

A stream with no duration makes `mpvLength` follow the position, so `Metadata` now changes once per second.


cmdrkotori:
Yeah, the way files stop has changed. IIRC we no longer navigate to a blank filename between files, so the check no longer works.

PR #1121, opened 2026-09-20
main: Prefer themed window icon on Unix

### LINKED ISSUE #1120: Wayland titlebar icon ignores themed desktop identity
## Environment

- **mpc-qt:** 26.07 Flatpak from Flathub (`io.github.mpc_qt.mpc-qt`)
- **Runtime:** `org.kde.Platform//6.11`
- **Desktop:** Fedora Kinoite, Plasma/KWin 6.7.5, native Wayland

## Observed behavior

The icon at the left side of the mpc-qt titlebar does not follow a user icon override on Plasma Wayland.

I have both of these local overrides:

- A local launcher desktop entry pointed at a preferred icon
- An app-ID-named theme override at `~/.local/share/icons/hicolor/scalable/apps/io.github.mpc_qt.mpc-qt.svg`

The launcher correctly shows the custom SVG. The titlebar continues to show mpc-qt's bundled icon. Closing/reopening the app and a cold reboot make no difference.

This is distinct from #337: the Flatpak desktop-file ID and exported icon name are already correct. The installed Flatpak exports:

```
hicolor/scalable/apps/io.github.mpc_qt.mpc-qt.svg
```

and `src/main.cpp` sets the matching desktop-file name, `io.github.mpc_qt.mpc-qt`.

## Code-level diagnosis

`src/main.cpp` unconditionally sets a bundled resource as the application icon:

```cpp
QApplication::setWindowIcon(QIcon(":/images/icon/mpc-qt.svg"));
```

This happens after `Flow::earlyPlatformOverride()` has set the desktop-file identity:

```cpp
QGuiApplication::setDesktopFileName("io.github.mpc_qt.mpc-qt");
```

In the Qt 6.11 QtWayland client used by the Flatpak, `QWaylandWindow::setWindowIcon()` stores the supplied icon locally and refreshes `QWaylandAbstractDecoration` when client-side decorations are active. Therefore, when Qt draws the titlebar, mpc-qt's fixed resource takes priority over the desktop identity and icon theme.

There is also a packaging inconsistency upstream: the source desktop entry uses `Icon=mpc-qt` and the Unix CMake install rule installs `mpc-qt.svg`, while the Flatpak already uses the more coherent app-ID name, `io.github.mpc_qt.mpc-qt`, for both its desktop-entry `Icon=` value and icon file.

## Suggested direction: one portable icon contract

Rather than a Wayland-only or KDE-specific exception, could mpc-qt make its desktop identity, themed Linux icon, and Qt runtime icon agree on one canonical name?

The proposed canonical Unix icon name is the existing desktop-file ID:

```
io.github.mpc_qt.mpc-qt
```

### Packaging

On Unix, install the icon under the canonical app-ID name and use that same name in the desktop entry:

```ini
# io.github.mpc_qt.mpc-qt.desktop
Icon=io.github.mpc_qt.mpc-qt
```

```text
share/icons/hicolor/scalable/apps/io.github.mpc_qt.mpc-qt.svg
```

Keeping `mpc-qt.svg` as a compatibility alias would avoid breaking existing distro packaging or user themes during a transition. Flathub already follows the canonical naming scheme.

### Runtime icon

On Unix, set Qt's runtime icon from the canonical theme name, with the current embedded SVG as fallback. On non-Unix platforms, preserve the current bundled runtime icon.

```cpp
const QIcon bundledIcon(":/images/icon/mpc-qt.svg");

#if defined(Q_OS_UNIX)
const QIcon runtimeIcon = QIcon::fromTheme(
    QStringLiteral("io.github.mpc_qt.mpc-qt"),
    bundledIcon
);
#else
const QIcon runtimeIcon = bundledIcon;
#endif

QApplication::setWindowIcon(runtimeIcon);
```

This is not a KDE path and does not need a Wayland-only branch:

- On Qt client-side decorations, the titlebar uses the same themed identity as the Linux desktop entry.
- On server-side decorations, taskbars, launchers, and titlebars continue to resolve `app_id` to the matching desktop entry and icon theme.
- On X11, Qt still has a valid runtime window icon.
- On Windows and macOS, the existing bundled runtime icon remains in place. Their executable/bundle icons are separate build-time artifacts.
- KDE, GNOME, wlroots compositors, and other Linux desktops all use the same freedesktop icon-theme contract.

The practical result is that the app's own client-side decoration no longer forces a different icon source than the rest of the desktop.

`xdg-toplevel-icon-v1` may eventually provide a standard per-window Wayland icon protocol, but that should be adopted by QtWayland rather than implemented directly by mpc-qt. It is not necessary for the identity/theme solution above.

## Scope of user overrides

A named icon-theme override is the portable Linux customization mechanism. An arbitrary absolute `Icon=/path/to/icon.svg` in a local desktop entry may not be visible to every sandboxed client-side decoration, so it cannot be guaranteed as an application-wide customization source. The app-ID-named hicolor override is available to the host Qt session and is the useful compatibility target here.

## Validation performed

- Confirmed `tweaksPreferWayland: true` in mpc-qt's Flatpak settings.
- Confirmed the Flatpak export and the in-app filesystem both contain the canonical `io.github.mpc_qt.mpc-qt.svg` icon.
- Confirmed the Flatpak's internal desktop entry uses `Icon=io.github.mpc_qt.mpc-qt` and matching `StartupWMClass`.
- Confirmed the local app-ID-named hicolor override exists.
- Confirmed, from the active Wayland session, that Qt resolves `QIcon::fromTheme("io.github.mpc_qt.mpc-qt")` successfully.
- Confirmed from the installed Qt 6.11 QtWayland library that `QWaylandWindow::setWindowIcon()` updates the local window icon and client decoration rather than sending a standard core-xdg-shell icon.
- Confirmed mpc-qt has no KDE Frameworks or Plasma-specific icon code path.

abirkel:
On Unix other than macOS, prefer the app-ID-themed icon for Qt window
decorations, with the embedded SVG retained as fallback.

Use `io.github.mpc_qt.mpc-qt` consistently as the desktop-entry icon name,
and install the SVG under that name while retaining `mpc-qt.svg` as a
compatibility alias.

This makes Qt client-side window decorations follow the same icon-theme
identity as the desktop entry, taskbar, and launcher.

Tested on Fedora Kinoite with Plasma Wayland:
- Custom `io.github.mpc_qt.mpc-qt` hicolor override appears in the titlebar.
- An isolated XDG environment falls back to the bundled mpc-qt icon.

Fixes #1120.


tbertels:
I think we also need to remove 
https://github.com/mpc-qt/mpc-qt/blob/f4195a217555e1e93fc29eb1952db3f6280f072a/packaging/flatpak/io.github.mpc_qt.mpc-qt.yml#L6
since the icon is already renamed during installation.

You can check by [enabling test builds in your fork](https://github.com/mpc-qt/mpc-qt/wiki/Contributing#enable-test-builds-in-your-fork) and pushing to your master branch, which will trigger a Flatpak build.
Then push with and without removing that line to make sure.

abirkel:
Removed `rename-icon` in ca76db28.

The fork Flatpak workflow completed successfully, and I installed the resulting
bundle on Fedora Kinoite with Plasma Wayland. The titlebar uses the custom
`io.github.mpc_qt.mpc-qt` themed icon as expected.

tbertels:
I hadn't noticed the cmake file edit actually adds a second copy of the icon, which is why the Flatpak build succeeds even with the rename in its manifest still in place.

I think it would be simpler to rename the "mpc-qt.svg" icon file in the repository and the code (there are still places where that name is used) and only use one icon named "io.github.mpc_qt.mpc-qt.svg".

abirkel:
I think I got them all. Please take a look.

Edit: Tested the updated flatpak build -- still working.

src/main.cpp
```diff
@@ -70,7 +71,12 @@ int main(int argc, char *argv[])
     Logger::setConsoleLogging(foundLoggingOpt);
     Logger::singleton();
     Logger::log(logModule, "starting logging");
-    QApplication::setWindowIcon(QIcon(":/images/icon/mpc-qt.svg"));
+    const QIcon bundledIcon(":/images/icon/io.github.mpc_qt.mpc-qt.svg");
```
tbertels: ":/images/icon/io.github.mpc_qt.mpc-qt.svg" should be a static constexpr char mpcQtIconPath[] like in mainwindow.cpp (or desktopFile in main.cpp).

src/propertieswindow.ui
```diff
@@ -15,7 +15,7 @@
   </property>
   <property name="windowIcon">
    <iconset>
-    <normalon>:/images/icon/mpc-qt.svg</normalon>
+    <normalon>:/images/icon/io.github.mpc_qt.mpc-qt.svg</normalon>
    </iconset>
   </property>
```
tbertels: That windowIcon property should be removed altogether, because it prevents that window from using the theme icon.

tbertels (changes requested):
The tray icon doesn't use the theme icon yet (see setupTrayIcon in mainwindow.cpp).
We manually create the icon in createIconFromSvg since d5eb21d345eeac38f80016dba0195fc8e266e6c2 but it doesn't seem needed anymore (tested with the AppImage, on Kubuntu 26.04 LTS, and on Linux Mint 22.2 Cinnamon).
So you can revert that commit.

abirkel:
Great catch. I don't use the tray and completely overlooked that. Thank you.

tbertels (approved):
It looks fine to me.

Could you rebase it to one commit?
And mention:
- The reason for this commit
- What it does (which is mostly the commits titles)
- That it reverts d5eb21d345eeac38f80016dba0195fc8e266e6c2
- The issue it fixes (including its title)

